PR #59, opened 2019-07-16
Merging hooks in one

munozemilio:


packages/cli/test/commands/configTelemetry.test.ts
```diff
@@ -38,20 +38,20 @@ describe('it should enable telemetry when a user opts in', () => {
...
 
-describe('it should take no action when telemetry is already enabled', () => {
-  test
-  .loadConfig({root: rootTelemetryEnabled})
-  .stdout()
-  .hook('init', {argv: ['arg']}, {root: rootTelemetryNull})
-  .do(output => expect(output.stdout).equal(''))
-  .it()
-})
+// describe('it should take no action when telemetry is already enabled', () => {
```
stevengum: if you use `xdescribe` the test will still show up in coverage but not run the actual test(s).

PR #92, opened 2019-08-15
added CLI form spec

scheyal:


scheyal:
We'll use readme's, generated and mini dev specs for alignment

PR #112, opened 2019-08-28
Making read file and read stdin data available through command class

jspruance:


packages/command/src/readpipeddata.ts
```diff
@@ -0,0 +1,36 @@
+const ReadPipedStdin = {
```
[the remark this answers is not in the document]
jspruance: unit test added

packages/command/src/readtextfile.ts
```diff
@@ -0,0 +1,49 @@
+const fs = require('fs-extra');
```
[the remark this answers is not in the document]
jspruance: unit test added

packages/chatdown/src/commands/chatdown.ts
```diff
@@ -74,16 +71,16 @@ export default class Chatdown extends Command {
...
     try {
-      // Check if path passed in --chat
+      // Check if path passed in --in
       if (args && args.length > 0) {
-        return txtfile.readSync(args)
+        return utils.readTextFile(args)
       } else {
         //Check if piped data was sent
         const {stdin} = process
         if (stdin.isTTY) {
```
munozemilio: Can we add this check to the utils as well?

packages/command/src/command.ts
```diff
@@ -53,6 +52,10 @@ export abstract class Command extends Base {
     process.stdin.destroy()
   }
 
+  readStdin() {
+    return ReadPipedData.read()
```
munozemilio: Should we try catch this and return false in case no input is passed?

jspruance: Added

packages/chatdown/src/commands/chatdown.ts
jspruance: This is done

Issue #98, opened 2019-08-19
Fix config:telemetry to follow pattern. bf config:show:telemetry


scheyal:
Specifically:
1. Need to show all configuration values. Use Show (not get)
2. Config:show displays all values whereas Config:show:telemetry shows only telemetry values

PR #128, opened 2019-09-13
Config commands standardization

munozemilio:
Fix for #98


packages/config/src/commands/config/set/telemetry.ts
```diff
@@ -0,0 +1,25 @@
...
+  static flags = {
+    help: flags.help({char: 'h'}),
+    disable: flags.boolean({char: 'd', description: 'Disable tlemetry'})
+  }
+
+  async run() {
+    const {flags} = this.parse(ConfigSetTelemetry)
+    let userConfig: Config = await getConfigFile(this.config.configDir)
+
+    if (flags.disable) {
```
axelsrz: consider changing userConfig.telemetry = !disable

PR #135, opened 2019-09-15
Added file/folder I/O CLI specs

scheyal:
Ignore changes to CLIFormSpec. I'd like to get it in, and I may just delete it later.
See changes to contrib guide: Force flag, and I/O processing rules


christopheranderson:
What does "Expected" mean in the table?

PR #139, opened 2019-09-18
[Do not merge] Add section support

feich-ms:
Add section support in lu file. Fixes https://github.com/microsoft/BotFramework-Composer/issues/733


packages/luis/parser/lufile/LUFileLexer.g4
```diff
@@ -1,7 +1,7 @@
 lexer grammar LUFileLexer;
 
 @lexer::members {
-  this.ignoreWS = true;             // usually we ignore whitespace, but inside template, whitespace is significant
+  ignoreWS = true;             // usually we ignore whitespace, but inside template, whitespace is significant
```
boydc2014: inside intent or inside untterance

packages/luis/parser/lufile/LUFileParser.g4
```diff
@@ -8,6 +8,7 @@ file
 
 paragraph
     : newline
+    | sectionDefinition
```
boydc2014: sectionDefintiion now contains intentDefinition? right?

feich-ms: not exactly. Intent can be outside section like 
# section
## intent1InSection
- xxx

# standaloneIntent
- xxx

packages/luis/src/parser/converters/luistoluconverter.js
```diff
@@ -48,7 +48,7 @@ module.exports = {
                     }
                     fileContent += NEWLINE + NEWLINE;
                 }
-                fileContent += '## ' + intent.intent.name + NEWLINE;
+                fileContent += '# ' + intent.intent.name + NEWLINE;
```
vishwacsena: We should chat. This to me seems like we will break back compat.

feich-ms:
The test coverage drop mainly comes from newly added CRUD operations in luParser.js. Will add more tests soon.

vishwacsena:
We can review this once 
- Tests are reverting to keep the `## intent` so we do not break back-compat
- Remove luParser.js CRUD operations from this PR. Make a separate PR with this + tests.

@feich-ms please update this PR once the changes are in.

feich-ms:
@vishwacsena 
1. I have reverted all the test cases that changed "##" to "#"..
2. The CRUD changes are also removed.
3. The changes in json files of test cases are just related to content order, not content itself
4. if '> !# @enableSections = true' is not added in the header, below content will be identified as two intents(AA and BB, AA is an intent without utterances)
   #AA 
   ##BB
   \- hello world

   if '> !# @enableSections = true' is added in the header, above content will be identified as section(AA) and sub intent(BB)
6. For the test coverage drop, I have no idea why it drops so much. I looked at the generated report, it is mainly caused by parseFileContents.js. But based on the report, my newly added functions or contents in parseFileContents.js are all covered by my newly added tests. so maybe @munozemilio can help to take a look?

vishwacsena:
@feich-ms We discussed this in BF-CLI review meeting today and decided to take these + other related changes for composer into a composer branch once we are locked down for ignite release (hopefully this week). So marking this as Do not merge. The only changes we want to take to master for now are related to Ignite deliverables.

feich-ms:
@vishwacsena, Ok, please let me know when the composer branch is ready. Thanks.

feich-ms:
Close this PR first, since all the composer related changes including section and section CRUD support are in this branch https://github.com/microsoft/botframework-cli/tree/composer now. When it is decided to go into the master, we can create a new PR then.

PR #156, opened 2019-09-24
Fix for ludown sort

munozemilio:


packages/luis/src/commands/qnamaker/translate.ts
```diff
@@ -0,0 +1,76 @@
...
+    translate_comments: flags.string({description: 'When set, machine translate comments found in .lu or .qna file'}),
+    translate_link_text: flags.string({description: 'When set, machine translate link description in .lu or .qna file'}),
+  }
+
+  /* tslint:disable:forin no-for-in*/
+  async run() {
+    try {
+      const {flags} = this.parse(QnamakerTranslate)
+      let inputStat = await fs.stat(flags.in)
+      let outputStat = flags.out ? await fs.stat(flags.out) : undefined
```
jspruance: minor thing, but generally better to set var to null, not explicitly to undefined

packages/luis/src/commands/qnamaker/convert.ts
```diff
@@ -19,24 +19,22 @@ export default class QnamakerConvert extends Command {
...
-      this.inputStat = await fs.stat(flags.in)
-      let isQnA = !this.inputStat.isFile() ? true : path.extname(flags.in) === '.lu'
+      let inputStat = await fs.stat(flags.in)
+      let isQnA = !inputStat.isFile() ? true : path.extname(flags.in) === '.lu'
 
       // Parse the object depending on the input
       let result: any
       if (isQnA) {
-        let luFiles = await this.getLuFiles(flags.in, flags.recurse)
+        let luFiles = await file.getLuFiles(flags.in, flags.recurse)
```
jspruance: better to use 'const' if var no reassigned later

packages/luis/src/commands/luis/translate.ts
```diff
@@ -1,107 +1,76 @@
...
-
+      let isLu = !inputStat.isFile() ? true : path.extname(flags.in) === '.lu'
       let result: any
-      let luFiles: any
       if (isLu) {
-        luFiles = await this.getLuFiles(flags.in, flags.recurse)
-        result = await luTranslator.translateLu(luFiles, flags.translatekey, flags.tgtlang, flags.srclang, flags.translate_comments, flags.translate_link_text)
+        let luFiles = await fileHelper.getLuFiles(flags.in, flags.recurse)
+        result = await luTranslator.translateLuFile(luFiles, flags.translatekey, flags.tgtlang, flags.srclang, flags.translate_comments, flags.translate_link_text)
+      } else {
```
jspruance: lots of args here - wonder if it would be better as an object

packages/luis/parser/translator/lutranslate.js
```diff
@@ -32,26 +42,29 @@ module.exports = {
...
     // Tokenize to_lang
-    let toLang = to_lang.split(/[, ]/g);
+    let toLang = to_lang.split(/[, ]/g)
     for (idx in toLang) {
         let tgt_lang = toLang[idx].trim();
         if (tgt_lang === '') continue;
         try {
-            parsedLocContent = await translateHelpers.parseAndTranslate(fileContent, translate_key, tgt_lang, src_lang, translate_comments, translate_link_text, false, undefined)
+            parsedLocContent = await translateHelpers.parseAndTranslate(luObject, translate_key, tgt_lang, src_lang, translate_comments, translate_link_text, false, undefined)
         } catch (err) {
```
jspruance: no need to pass undefined here - just leave it blank and if no arg is passed it will be undefined by default

PR #164, opened 2019-09-26
Check for version updates no more than 1x per day, based on timestamp in config

jspruance:


packages/cli/src/hooks/init/inithook.ts
```diff
@@ -7,13 +7,54 @@ const latestVersion = require('latest-version')
...
+    }
+
+    const isToday = (dateObj: Date, today: Date) => {
+      return dateObj.getDate() === today.getDate() &&
+        dateObj.getMonth() === today.getMonth() &&
+        dateObj.getFullYear() === today.getFullYear()
+    }
+
+    // if there's no timestamp in config, create one and check for updates
+    if (!userConfig.lastVersionCheck) {
```
munozemilio: Can this be simplified to avoid duplicated code? e.g
    const lastCheck = userConfig.lastVersionCheck ? new Date(userConfig.lastVersionCheck) : Yesterday
    if (!isToday(lastCheck, curDateTime)) {
          await checkForUpdate()
           updateUserConfig(curDateTime)
    }

packages/cli/src/hooks/init/inithook.ts
```diff
@@ -50,7 +82,7 @@ const hook: Hook<'init'> = async function (opts) {
 
       await fs.mkdirp(this.config.configDir)
 
-      await fs.writeFileSync(path.join(this.config.configDir, 'config.json'), JSON.stringify(userConfig, null, 2))
+      fs.writeFileSync(path.join(this.config.configDir, 'config.json'), JSON.stringify(userConfig, null, 2))
```
munozemilio: Can we switch this to use the async version and use writeFile?

packages/cli/src/hooks/init/inithook.ts
jspruance: great suggestion...updated

packages/cli/src/hooks/init/inithook.ts
jspruance: this is updated

packages/cli/src/hooks/init/inithook.ts
```diff
@@ -7,14 +7,51 @@ const latestVersion = require('latest-version')
...
+      if (semver.gt(latest, opts.config.version)) {
+        this.log('Update available ')
+        this.log('     Run ')
+        this.log(`npm i -g ${opts.config.name} `)
+      }
+    }
+
+    const updateUserConfig = (curVersionCheck: Date) => {
+      userConfig.lastVersionCheck = curVersionCheck
+      fs.writeFileSync(path.join(this.config.configDir, 'config.json'), JSON.stringify(userConfig, null, 2))
```
munozemilio: we need to ensure the path exists to avoid any writing error in case the user deletes the folder o is inexistent,
we can add await fs.mkdirp(this.config.configDir) before line 38, or even better extract this to a function like writeConfig(userConfig) and call it from here and from line 79

jspruance: this is complete

jspruance: updated PR with file writing refactor & add lastVersionCheck during post install

PR #172, opened 2019-10-01
.lu file update - Ability to tie features to a specific model

vishwacsena:
- [x] Added new tests
- [x] Verified all existing tests pass

## Tie features to a specific model
Phrase lists today apply to all models in an application. LUIS is changing this to provide users more fine grained control where users can explicitly assign a phrase list as a feature to another model. With this, phrase list can be added as a feature to 

- another intent
- another entity

Here's how you define phrase list as a feature to another model

```markdown
> phrase list definition
@ phraseList city(interchangeable) = 
    - seattle
    - space needle
    - SEATAC
    - SEA

> phrase list as feature to intent (also applicable to entities)
@ intent getUserProfileIntent usesFeature city

> Other entity types can also use this feature
@ simple xyz usesFeature city
```


packages/luis/src/parser/converters/luistoluconverter.js
```diff
@@ -98,8 +98,16 @@ module.exports = {
                     fileContent += NEWLINE + NEWLINE;
                 }
                 fileContent += `@ simple ${entity.name}`;
-                if (entity.roles.length > 0) {
-                    fileContent += ` ${entity.roles.length > 1 ? `hasRoles` : `hasRole`} ${entity.roles.join(',')}`
+                if (entity.roles && entity.roles.length > 0) {
+                    fileContent += ` ${entity.roles.length > 1 ? `hasRoles` : `hasRole`} ${entity.roles.join(',')}`;
+                }
+                if (entity.features && entity.features.length > 0) {
+                    let featuresList = new Array();
```
munozemilio: Is there a way top extract this in a function to avoid duplicated code?

vishwacsena: @munozemilio - done. Please review.

PR #175, opened 2019-10-01
Version 1 of contact.form.dialog

veritas2019:
* Same info is modeled for both Home and Work.
* List of state abbreviations and their spelled-out names are specified in the separate state.en-us.lu, and used in the $mappings specifications.
* Latitude and Longitude should be auto-filled if the corresponding address are filled.
* Validation patterns are specified for zip codes and phone numbers.

-- FIXES
* contact.form.dialog
  - Fixed spelling
  - Use phonenumber prebuilt entity type
* activity.form.dialog: the second scenario from Darren "An employee working for a customer wants to add a quick update to a customer record after a call. This update will then be available for other people working with that customer to see".

### DIFF
--- a/packages/dialog/templates/en-us/state.en-us.lu
+++ b/packages/dialog/templates/en-us/state.en-us.lu
@@ -1,0 +1,110 @@
+**ENUMS
+$STATE:AL=
+-Alabama
+$STATE:AK=
+-Alaska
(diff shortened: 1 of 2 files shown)


veritas2019:
Screenshot from Darren's email about the activity scenario

I suspect more needs to be added for this?

PR #177, opened 2019-10-02
Model as feature to other models - support in lu format

vishwacsena:
## Model as feature

With this, models (all intents and entity types) can be used as a feature to another model (intent and entities). 

Constraints: 
1. Cirular model as feature is not supported. e.g. A isFeatureTo B and B isFeatureTo A. 
2. Both source model and target model should be defined and exist in the application.
3. An entity, phrase list and intent all can have the same name. So when an ambiguous definition appears on either the left or ride side of `usesFeature`, we should throw a parsing error that calls out the ambiguity. 
    E.g. user could have intent named `city` and an entity named `city` as well as a phrase list named `city`. If user tries to do @ city usesFeature geographyV2, `city` in this definition is ambiguous since it is not clear which specific `city` user would like to add `geographyV2` as a feature. In this case, the parser should throw an error and state `city` is ambiguous. Found an intent, phrase list and entity with that name. Consider renaming the overlapping concepts to establish a usesFeature relationship. 

Here's how you define entity as a feature to another entity - with `usesFeature`.

```markdown
> entity definition - <> \<entityType> \<entityName> [<roles>]
@ prebuilt personName
@ prebuilt age

> entity definition with roles
@ simple userName fistName, lastName

> add entity as feature to another entity
@ userName usesFeature personName

> add entity as feature to intent
@ getUserNameIntent usesFeature personName

> Intent definition
# getUserNameIntent
- utterances

> entity as feature to multiple models - in this case another entity and an intent
@ userName usesFeature personName
@ intent getUserNameIntent usesFeature personName

> multiple entities as feature to a model
@ getUserNameIntent usesFeature age, personName

> intent as feature to another intent
@ intent getUserProfileIntent usesFeature getUserNameIntent

# getUserProfileIntent
- utterances
```


vishwacsena:
@munozemilio notice the improvement to test coverage :)

PR #178, opened 2019-10-02
Removing Disk I/O operatios as mandatory for lu transform

munozemilio:


packages/luis/src/parser/converters/lumerger.js
```diff
@@ -16,77 +16,53 @@ const luisJSON = require('./../luisfile/parseLuisFile');
...
-                    file.filePath = path.resolve(parentFilePath, file.filePath);
-                } 
-                // avoid parsing files that have been parsed already
-                if(parsedFiles.includes(file.filePath)) {
-                    // find matching parsed files and ensure includeInCollate is updated if needed.
-                    updateParsedFiles(allParsedLUISContent, allParsedQnAContent, allParsedAlterationsContent, file);
-                } else {
-                    filesToParse.push(new filesToParseClass(file.filePath, file.includeInCollate));
-                }
+        let parentDir = luOb.id === 'stdin' ? process.cwd() : path.parse(path.resolve(luOb.id)).dir
```
vishwacsena: Does this not mean we cannot support wild cards in references? That's what the previous logic was supporting.

It might be cleaner for us to not provide a parentDir to the luSearchFn and instead just provide the ID of the thing that needs resolution. If this is a relative path, then the ID will be the relative path itself and the luSearchFn should know how to handle it. If this was a glob pattern, then we just pass the glob pattern up and have the delegate resolve.

PR #185, opened 2019-10-08
Luis commands stdin enabling

munozemilio:

PR #186, opened 2019-10-08
Stdin on QnAMaker

munozemilio:


vishwacsena:
We should get #185 in and FI from master so it would be easy to just see the delta from that PR.

vishwacsena:
@munozemilio - this now has merge conflicts with master. Can you help resolve?

PR #189, opened 2019-10-08
Don't throw an error when we intentionally exit with exit code 0

jspruance:
When we issue a command with no flags (ex: 'bf chatdown') the help contents are displayed.

The oclif base command then returns an error with exit code 0, which chatdown then throws in the catch clause:

exit(code = 0) { 
  return Errors.exit(code); 
}

Since exit code 0 is an intentional way to exit the Node program, I believe we should not be throwing an error.

This change looks for 'EEXIT: 0' in the error message returned from oclif and if present exits gracefully with 'process.exit(0)'.


munozemilio:
I will submit a PR to handle this case in the command base class

jspruance:
> I will submit a PR to handle this case in the command base class

yes that would be better. then we don't have to make changes to every command

PR #192, opened 2019-10-10
Nigao/confirmation

ninggao:
(1) contextual help
(2) revised schemas from Ben

### DIFF
--- a/packages/dialog/templates/en-us/common.en-us.lg
+++ b/packages/dialog/templates/en-us/common.en-us.lg
@@ -24,6 +24,9 @@
 # helpProperty(property, name)
 - Possible values for {name} include: \[{join(turn.schema.properties[property].enum, ',')}\]
 
+# contextHelp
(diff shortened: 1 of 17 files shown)


packages/dialog/templates/en-us/help.en-us.lu
```diff
@@ -2,4 +2,7 @@
 # Help**PROPERTY**
```
chrimc62: Thought you were going to remove help on specific intents?

packages/dialog/templates/en-us/common.en-us.lg
```diff
@@ -24,6 +24,9 @@
 # helpProperty(property, name)
 - Possible values for {name} include: \[{join(turn.schema.properties[property].enum, ',')}\]
```
chrimc62: Let me get synced so we can compute and call the right lg template.

PR #193, opened 2019-10-11
nDepth entity support in lu format

vishwacsena:
## [New] Simple entity with n-depth support

Here's a definition of an `address` n-depth entity with `fromAddress` and `toAddress` as two roles.

```markdown
@ list listCity
@ prebuilt number
@ prebuilt geographyV2
@ regex regexZipcode = /[0-9]{5}/
@ ml address hasRoles fromAddress, toAddress 
@ address =
    - @ number 'door number'
    - @ ml streetName
    - @ ml location usesFeature geographyV2
        - @ listCity city
        - @ regexZipcode zipcode
```


packages/luis/src/parser/converters/luistoluconverter.js
```diff
@@ -170,15 +175,58 @@ module.exports = {
...
+ */
+const addNDepthChildDefinitions = function(childCollection, tabStop, fileContent) {
+    let myFileContent = '';
+    (childCollection || []).forEach(child => {
+        myFileContent += "".padStart(tabStop * 4, ' ');
+        myFileContent += '- @ ';
+        if (child.instanceOf) {
+            myFileContent += child.instanceOf;
+        } else if (child.children && child.children.length !== 0){
+            myFileContent += EntityTypeEnum.ML;
```
munozemilio: this if does exactly the same as the else condition, we can remove the else if.

packages/luis/src/parser/index.js
```diff
@@ -4,7 +4,6 @@
  */
 const modules = {
     parser: {
-        parseLuList: require('./converters/lutoluisconverter').parseLuToLuis,
```
munozemilio: Should we expose this for composer?

packages/luis/src/parser/lufile/parseFileContents.js
```diff
@@ -152,7 +152,76 @@ const parseLuAndQnaWithAntlr = async function (parsedContent, fileContent, log,
...
+                        child.features[idx] = new helperClass.modelToFeature(feature);
+                        featureHandled = true;
+                    }
+                }
+                if (!featureHandled) {
+                    // find feature as intent
+                    let intentFeatureExists = intentsCollection.find(i => i.name == feature);
+                    if (intentFeatureExists) {
+                        child.features[idx] = new helperClass.modelToFeature(feature);
+                        featureHandled = true;
```
munozemilio: if this condition is valid featureHanlded is already true

packages/luis/src/parser/lufile/parseFileContents.js
```diff
@@ -152,7 +152,76 @@ const parseLuAndQnaWithAntlr = async function (parsedContent, fileContent, log,
...
+                    }
+                }
+                if (!featureHandled) {
```
munozemilio: can this be an else condition of the previous validation?

packages/luis/src/parser/lufile/parseFileContents.js
```diff
@@ -891,7 +944,131 @@ const parseAndHandleEntityV2 = function (parsedContent, luResource, log, locale)
...
+            throw (new exception(retCode.errorCode.INVALID_INPUT, errorMsg));
+        }
+        let context = {line : defLine, definition: child.trim()};
+        switch(groupsFound.groups.instanceOf.toLowerCase().trim()) {
+            case EntityTypeEnum.SIMPLE:
+                if (!currentParentEntity.entity.children) {
+                    currentParentEntity.entity.children = new Array(new helperClass.childEntity(childEntityName,"", context, [], childFeatures));
+                } else {
+                    // de-dupe and push this child entity    
+                    let childExists = (currentParentEntity.entity.children || []).find(item => item.name == childEntityName);
```
munozemilio: this code appears 3 times, can we extract to a function?

packages/luis/src/parser/lufile/parseFileContents.js
```diff
@@ -891,7 +944,131 @@ const parseAndHandleEntityV2 = function (parsedContent, luResource, log, locale)
...
+    });
+    if (entityFound !== undefined) {
+        let errorMsg = `${matchType} Prior definition - '@ ${entityFound.type} ${entityFound.name}${entityFound.roles.length > 0 ? ` hasRoles ${entityFound.roles.join(',')}` : ``}'`;
+        let error = BuildDiagnostic({
+            message: errorMsg,
+            context: line
+        })
+        throw (new exception(retCode.errorCode.INVALID_INPUT, error.toString()));
+    }
+    return {matchType, entityFound}
```
munozemilio: why return this if only validation is required?

packages/luis/src/parser/index.js
vishwacsena: Not for now. Because once we do this we will need to keep this since it will be a public API. We can expose what we need for composer in the composer branch for now; stabilize the API and then make it public.

packages/luis/src/parser/lufile/parseFileContents.js
```diff
@@ -891,7 +943,121 @@ const parseAndHandleEntityV2 = function (parsedContent, luResource, log, locale)
...
+        } 
+        
+        currentParentEntity = entityIdxByLevel.reverse().find(item => item.level == tabLevel - baseTabLevel);
+        entityIdxByLevel.reverse();
+        if (!currentParentEntity) {
+            let errorMsg = `[ERROR] line ${defLine}: Invalid definition found for child "${child.trim()}". Parent of each child entity must be of type "${EntityTypeEnum.ML}".`;
+            throw (new exception(retCode.errorCode.INVALID_INPUT, errorMsg));
+        }
+        let context = {line : defLine, definition: child.trim()};
+        if (groupsFound.groups.instanceOf.toLowerCase().trim() === EntityTypeEnum.SIMPLE) {
```
munozemilio: I think we can simplify like this
        let childET
        if (groupsFound.groups.instanceOf.toLowerCase().trim() === EntityTypeEnum.ML) {
            let newParent = currentParentEntity.entity.children.find(item => item.name == childEntityName);
            // Push the ID of the parent since we are proessing the first child entity
            entityIdxByLevel.push({level : tabLevel - baseTabLevel + 1, entity : newParent});
        } else {
            childET = childEntityType
        }

        pushNDepthChild(currentParentEntity.entity, childEntityName, context, childFeatures, childET);

vishwacsena: The find line is operating on the collection and the push needs to happen before the find.

PR #196, opened 2019-10-11
Further refactor of chatdown and fix 1 bug

jspruance:
- Write to output dir (--out) was only working when a dir passed in (--in) but not for single files. This is fixed.


packages/chatdown/src/commands/chatdown.ts
```diff
@@ -37,12 +37,14 @@ export default class Chatdown extends Command {
         })
       }
 
+      let outputDir = (flags.out) ? flags.out.trim() : './'
+      if (outputDir.substr(0, 2) === './') {
+        outputDir = path.resolve(process.cwd(), outputDir.substr(2))
```
munozemilio: let outputDir = path.resolve(flags.out) should return the same result

packages/chatdown/src/commands/chatdown.ts
```diff
@@ -91,21 +94,27 @@ export default class Chatdown extends Command {
...
+  private getFileName(file: any) {
+    let fileName = file
+    if (file.lastIndexOf('/') !== -1) {
+      fileName = fileName.substr(fileName.lastIndexOf('/') + 1)
+    }
+    if (file.lastIndexOf('\\') !== -1) {
+      fileName = fileName.substr(fileName.lastIndexOf('\\') + 1)
+    }
+    fileName = fileName.substring(0, fileName.lastIndexOf('.'))
+    return fileName
```
munozemilio: let fileName = path.basename(file, path.extname(file)) probably does the same

packages/chatdown/src/commands/chatdown.ts
```diff
@@ -119,7 +128,13 @@ export default class Chatdown extends Command {
     })
   }
 
-  private async writeOut(activities: any) {
+  private async writeOut(activities: any, fileName: string, outputDir: any) {
+    if (fileName && outputDir) {
+      let writeFile = `${outputDir}\\${fileName}.transcript`
```
munozemilio: let writeFile = path.join(outputDir, ${fileName}.transcript)

packages/chatdown/src/commands/chatdown.ts
jspruance: Updated old lib code to use path library

packages/chatdown/src/commands/chatdown.ts
jspruance: updated

packages/chatdown/src/commands/chatdown.ts
jspruance: updated

packages/chatdown/src/commands/chatdown.ts
```diff
@@ -37,12 +37,15 @@ export default class Chatdown extends Command {
         })
       }
 
+      // let outputDir = (flags.out) ? flags.out.trim() : './'
```
munozemilio: Can we remove comments?

packages/chatdown/src/commands/chatdown.ts
```diff
@@ -37,12 +37,15 @@ export default class Chatdown extends Command {
...
+      // if (outputDir.substr(0, 2) === './') {
+      //   outputDir = path.resolve(process.cwd(), outputDir.substr(2))
+      // }
+      let outputDir = flags.out ? path.resolve(flags.out) : path.resolve(process.cwd())
```
munozemilio: if flags.out is empty should we print to console?

packages/chatdown/src/commands/chatdown.ts
jspruance: oops, removed!

packages/chatdown/src/commands/chatdown.ts
jspruance: i don't think so, since 'out' is totally optional

PR #199, opened 2019-10-13
doc updates - ludown

vishwacsena:


vishwacsena:
@munozemilio - are the changes to readme.md right? I thought those were auto-generated.. ? If so, how do we ensure that the links to the file formats are preserved?

PR #214, opened 2019-10-16
Rename of luis folder to lu

munozemilio:


munozemilio:
@vishwacsena, @Danieladu @feich-ms FYI

PR #218, opened 2019-10-17
Allow scanning of dir without blob expression

jspruance:


jspruance:
Code was only successfully reading files from dir when a glob pattern was in the path. This fixes that.

PR #219, opened 2019-10-18
nDepth entity with features

vishwacsena:
- Fixed bug where prebuilt entity should be `phonenumber` instead of `phoneNumber`
- nDepth entities in labelled utterances are now handled correctly
    - duplicates are removed from `entities` collection
    - library throws if parent of a nDepth child is not labelled in the utterance.
- Added new tests
- Verified E2E works via LU -> JSON -> manual import new app in staging.


packages/lu/src/parser/converters/luistoluconverter.js
```diff
@@ -235,6 +250,7 @@ const addNDepthChildDefinitions = function(childCollection, tabStop, fileContent
         myFileContent += NEWLINE;
         if (child.children && child.children.length !== 0) {
             myFileContent += addNDepthChildDefinitions(child.children, ++tabStop, myFileContent);
```
munozemilio: tabStop +1 ?

Issue #247, opened 2019-10-18
[chatdown][luis][qnamaker] missing force flag


christopheranderson:
## Problem:

Any command with `--out` must support `--force` to be compliant to spec. The following commands that have `--out` commands do not support `--force:

- `bf chatdown`
- `bf luis:convert`
- `bf luis:translate`
- `bf qnamaker:convert`
- `bf qnamaker:translate`

## Desired solution

Ensure that these commands do take a `--force` flag and that it doesn't overwrite files unless that flag is passed.

scheyal:
Add for consistency, yes.

Issue #246, opened 2019-10-18
[qnamaker] qnamaker:export:kb doesn't allow me to choose the file


christopheranderson:
## Problem:

I'd expect that I can choose the output given the text description:

```
λ  npx bf qnamaker:export:kb --help
Export a knowledgebase to .json file
```

But the options don't allow `--out`

```
OPTIONS
  -h, --help                         qnamaker:export:kb command help
  --endpoint=endpoint                Overrides public endpoint https://westus.api.cognitive.microsoft.com/qnamaker/v4.0/
  --environment=environment          (required) Specifies whether environment is Test or Prod.
  --kbId=kbId                        Knowledgebase id to be exported. Overrides the knowledge base id present in the config
  --subscriptionKey=subscriptionKey  Specifies the qnamaker Ocp-Apim-Subscription Key (found in Keys under Resource Management section for your Qna Maker cognitive service). Overrides the subscriptionkey value present in the config
```

## Desired solution

Either change the doc to indicate it echos json to stdout 

OR

Add `--out` support.

Issue #241, opened 2019-10-18
[Hackathon] Check for redundancies like config/src/commands/index.ts


v-kydela:
The files [config/src/commands/index.ts](https://github.com/microsoft/botframework-cli/blob/master/packages/config/src/commands/index.ts) and [config/src/commands/config/index.ts](https://github.com/microsoft/botframework-cli/blob/master/packages/config/src/commands/config/index.ts) are almost the same. Are both of these files supposed to exist? Do they serve distinct purposes?

Issue #239, opened 2019-10-18
[all] support short form options for common options


christopheranderson:
## Problem:

All options currently support only long form (aka `--`) options. Many clis out there support short form options for things that are common. (i.e. `git checkout -b` instead of `git checkout --branch`).

## Desired Behavior:

For frequently used options (`help`, `in`, `out`, etc.), it would be great if we supported short hand options like `-h`, `-i`, `-o`.

PR #252, opened 2019-10-19
Luis command handle empty lu file

### LINKED ISSUE #230: Empty .lu file creates error on luis:convert
Problem:

When using an empty .lu file, luis:convert errors out.

```
Unknown error during execution. Please file an issue on https://github.com/microsoft/botframework-cli/issues
Cannot destructure property `fileContent` of 'undefined' or 'null'.
```

Repro steps:

1. Create an empty Main.lu file
2. run `bf luis:convert--in ./Main.lu`
3. Get error

Desired behavior:

It should handle empty files.

munozemilio:
Fixes #230, #241, #247, #239, #246


packages/lu/src/parser/converters/lumerger.js
```diff
@@ -327,7 +327,11 @@ const readLuFile = function(file) {
 }
 
 const parseLuFile = async function(file, log, luis_culture) {
-    let parsedContent = '';
+    let parsedContent = ''
+    if (!file) {
+        let error = BuildDiagnostic({ message: `Sorry, empty file content` })
```
christopheranderson: [blocking] All error messages should provide guidance to users on how to resolve them.

@vishwacsena - is erroring on empty `.lu` the correct behavior here? If so, what action should the user take.

Alternatively, I feel like a warning with no-op might be a better pattern, especially when passed in a directory reference.

christopheranderson: Talked offline. We're going to go with error. Suggested error text:

> Cannot parse empty ${filename}. Please add content to the file or remove it.

packages/qnamaker/src/commands/qnamaker/kb/export.ts
```diff
@@ -4,7 +4,7 @@ const exportKbJSON = require('./../../../../utils/payloads/exportkb')
 import {Inputs, processInputs} from '../../../utils/qnamakerbase'
 
 export default class QnamakerKbExport extends Command {
-  static description = 'Export a knowledgebase to .json file'
+  static description = 'Echos a knowledgebase .json to stdout'
```
christopheranderson:
```suggestion
  static description = 'Echos a knowledgebase json to stdout'
```

[nit] `.json` is a file extenion. `json` is a data format.

PR #276, opened 2019-10-23
Update config plugin help flags

munozemilio:


packages/chatdown/README.md
```diff
@@ -14,7 +14,7 @@ Tool for parsing chat files and outputting replayable activities
 
 ## `bf chatdown`
 
-Converts chat dialog files in <filename>.chat format into transcript files. Writes corresponding <filename>.transcript for each .chat file.
+Converts chat dialog files in <filename>.chat format into transcript file. Writes corresponding <filename>.transcript for each .chat file
```
christopheranderson: Why is it `file` here and `files on line 31?

christopheranderson: Also removed a period here, but you added it in other places.

PR #283, opened 2019-10-28
Power BI usage stats dashboard file

scheyal:


scheyal:
snicked readme file. re-doing.

PR #285, opened 2019-10-29
dialog cli spec

scheyal:


README.md
```diff
@@ -1,100 +1,100 @@
-
```
scheyal: This should be identical to main. I just can't remove it from the commit/merge.

PR #295, opened 2019-10-29
Fixing output file path messages

munozemilio:


packages/config/src/commands/config/index.ts
```diff
@@ -4,7 +4,7 @@
  */
 
 import {Command, flags} from '@oclif/command'
-
+const path = require('path')
```
christopheranderson: nit: why not use import?

PR #298, opened 2019-10-30
Lu parser re architecture to be OO

munozemilio:
The main idea is to model the parser in a way we can create Lu, Luis, QnA and Alterations objects.
Lu:
```
class Lu {
    constructor(content, id, includeInCollate = true, luis = null, qna = null, alterations = null)

     static async build(luArray)

     async parseToLuis()

     async parseToQna()

     async parseToQnaAlterations()
}
```

Luis:
```
class Luis {
    constructor(luisJson)

     static build(luisJsonArray)

      parseToLuContent()

     sort()

     validate()
}
```

QnA:
```
class Qna {
    constructor(qnaJson)

     static build(qnaJsonArray)

      parseToLuContent()

     sort()

     validate()
}
```

This will be more intuitive for developers who will consume this as a library, easier to maintain, and with more clear responsibility at class level. eg


packages/lu/src/paarser/luis/luis.js
```diff
@@ -0,0 +1,789 @@
...
+        this.entities.sort(sortComparers.compareNameFn);
+        this.closedLists.sort(sortComparers.compareNameFn);
+        this.regex_entities.sort(sortComparers.compareNameFn);
+        this.model_features.sort(sortComparers.compareNameFn);
+        this.patternAnyEntities.sort(sortComparers.compareNameFn);
+        this.prebuiltEntities.sort(sortComparers.compareNameFn);
+        this.utterances.sort(sortComparers.compareIntentFn);
+    }
+
+    static build(parsedLUISList) {
```
boydc2014: Here and the `build` method in qna all take an array of json, and `merge` into one. 

Which sounds like we should put this into luMerger? or some Merger, if we are using the merger 'abstraction'

packages/lu/src/paarser/lu/lu.js
```diff
@@ -0,0 +1,62 @@
...
+
+    static async build(luArray){
+        let mergedContent = await mergeLuFiles(luArray)
+        let luisContent = mergedContent.LUISContent.filter(item => item.includeInCollate)
+        let qnaContent = mergedContent.QnAContent.filter(item => item.includeInCollate)
+        let qnaAlterationsContent = mergedContent.QnAAlterations.filter(item => item.includeInCollate)
+        let luisResult = Luis.build(luisContent)
+        let qnaResult = QnA.build(qnaContent)
+        let qnaAlterationsResult = Alterations.build(qnaAlterationsContent)
+        let luContent = '' + luisResult.parseToLu()
```
boydc2014: We are converting the parsed content (json) back to raw content? is this necessary?

packages/lu/src/paarser/lu/lu.js
```diff
@@ -0,0 +1,62 @@
...
+        }
+        return this.alterations   
+    }
+}
+
+module.exports = Lu
+
+const updateParsedObjects = async function(lu){
+    let parsedContent = await parseFileContents(lu.content)
+    lu.luis = new Luis(parsedContent.LUISJsonStructure)
```
boydc2014: I thought lu.luis, lu.qna, lu.alternations are already set in the build function. When this function could be called? 

and generally speaking, if we create a static `build` method, which seems like a typical Factory pattern, which usually the constructor is private. So it's a little bit odd that we call the constructor directly here.  Move the `build` method out of Luis class into a new `LuisMerge` class probably help solve this kind of smell.

packages/lu/src/paarser/lu/lu.js
```diff
@@ -0,0 +1,62 @@
...
+        let qnaAlterationsResult = Alterations.build(qnaAlterationsContent)
+        let luContent = '' + luisResult.parseToLu()
+        luContent += qnaResult.parseToLu()
+        luContent += qnaAlterationsResult.parseToLu()
+        let result = new Lu(luContent, '', true, luisResult, qnaResult, qnaAlterationsResult)
+        return result
+
+ }
+
+    async parseToLuis(){
```
boydc2014: If those field are already be set in `build`, in what scenario, this would be null and set through this way? 

I mean, we should have a single entry of this function\module.  I assume it's this `build` function (it don't have to be), then we don't want to deal with the case that the luis, qna, alternation maybe not be there. 

If we think build is not the single entry, we should move build out, and make the constructor a single entry, and don't allow build to directly set property like `lu.Luis = xxx`.  We either do it in the constructor. or leave this "lazy-initialization" way, it's both OK, as long as there is only one way to construct such a object.

packages/lu/src/paarser/luis/luis.js
munozemilio: Yes actually this comes from the lumerger. Since thats the simplest way to merge all the lu references into one. Ill think into a different way to abstract this then

packages/lu/src/paarser/lu/lu.js
munozemilio: I just though to add it since the idea is to build a Lu object from multiple Lu files referencing each other

packages/lu/src/paarser/lu/lu.js
munozemilio: The reason for this is when you create an instance of the Lu object passing the content of a single file (e.g when reading the content from stdin in cli). In this scenario it is easier to initialize the object directly and not wrapping it in an array. Agree on moving to another file though

packages/lu/src/paarser/lu/lu.js
munozemilio: Agree. Ill just create a builder class

packages/lu/src/paarser/lu/lu.js
boydc2014: Yes, i think you do have scenarios where you want to new this class directly. The point is in this case and any other cases, it would be great to have one way to construct such a class. And it should be doable

So we are supporting two workflow here
1. someone use the `build` method to merge, parse, `merge` again
2. someone directly new a constructor, and we do lazy parsing when they want to get a property. 

Sounds like 2 can be included into 1, because we can and should wrap any files from stdin, also into a luObject. luObject will be our single entry point.

packages/lu/src/paarser/lu/lu.js
boydc2014: My original point is given this luContent is already not the original content, how useful it would be in some scenarios. 

If there are certain scenarios depends on that, it can be implemented in a "lazy" way like parseToLu() in it's children properties, for a consistency perspective.  Which means, either we, you know, inside the Luis class, we also do the same thing to get the flushed content (cache the content), and in there, `luContent = luisResult.content + ...content` something like this.  Or, we implement a method here parseToLu() in which we call luisResult.parseToLu() + other.pasreToLu().

You see, either content = content + content .. ,  or parseToLu = parseToLu + parseToLu + parseToLu. Bear with my poor English. :)

packages/lu/src/paarser/lu/lu.js
boydc2014: Cool, replied under another comment that the point is single entry which is LuObject.  Covert to luObject as early as possible in the process pipeline, will make the following steps in the pipeline have `less state`, be more focus, do one thing and do it well.

vishwacsena:
I'll leave a general top level comment here - 

1. We should clean up the code base such that 
    a. .lu file describe LU content 
    b. .qna file describe QnA content. We already started this clean up via docs and samples as well as top level recommendation but should go to the extent of enforcing/ requiring it to keep things cleaner.
2. For both .lu and .qna, we should support an API that can take one or more .lu/ .qna files and return one LUIS or one QnA model. For QnA we should also include an alterations model for any alterations information found in the .qna file(s).
 
I believe the core issue we have today is two-fold

`parsedContent` is a clobbered in memory object that has LUIS, QnA, QnA alterations and every other random house keeping that is being tracked there. We should clean this up. 

The core parseFileContent.js as well as collate function is getting out of control since there are multiple walks over respective LUIS and QnA objects in parsedContent. This should be cleaned up by having a LUIS, QnA, QnAContent classes that has a static parseContent method that can take an (a collection of .lu or .qna content) and returns a new instance of LUIS or QnA or QnAContent. We should also include a collate() method on these classes that can take additional content (.lu or .qna) and collate the content internally within that object. And finally, include a validate method that can run necessary cross validations that consumer can call after all parseContent/ collate calls. 

This will help cleanup the current code substantially and make things more organized and self-contained. 

From a pure consumption as a library POV, I would expect something like this to work - 

```
var contentToParse = [new luObj(`id`, `>lu content`)];
var newContentToCollate = [new luObj(`id`, `> more lu content`)];
var parsedLUISObj = LuToLUIS.parseContent(contentToParse);
parsedLUISObj.collate(newContentToCollate);
parsedLUISObj.validate()

// to consume
parsedLUISObj.intents
parsedLUISObj.entities
...
```

@munozemilio - if these resonate as the core problems we are trying to solve, I would recommend that we lead with those and call out the core outcomes we would like to achieve here. If not,

packages/lu/src/parser/translator/lutranslate.js
```diff
@@ -1,38 +1,43 @@
...
-                throw(err);
-            }
-         }
-        return translation
+        return await translateMarkDownList(files, translate_key, to_lang, src_lang, translate_comments, translate_link_text, true)
     },
-    translateLuObj: async function(luObject, translate_key, to_lang, src_lang, translate_comments, translate_link_text) {
-        let translation = {}
+    translateLu: async function(luObject, translate_key, to_lang, src_lang, translate_comments, translate_link_text) {
+        return await translateMarkDown(luObject, translate_key, to_lang, src_lang, translate_comments, translate_link_text, true)
```
christopheranderson: avoid `return await` unless you're in a `try-catch`

packages/lu/src/parser/translator/lutranslate.js
```diff
@@ -1,38 +1,43 @@
...
+    translateQnA: async function(qnaObject, translate_key, to_lang, src_lang, translate_comments, translate_link_text) {
+        return await translateMarkDown(qnaObject, translate_key, to_lang, src_lang, translate_comments, translate_link_text, false)
+    }
+}
+
+
+const translateMarkDownList = async function(files, translate_key, to_lang, src_lang, translate_comments, translate_link_text, isLu) {
+    let translation = {}
+    let i = 0
+    while(files.length > i) {
```
christopheranderson: switch to for-of?

packages/lu/test/parser/lufile/parseFileContents.composeEntity.test.js
```diff
@@ -62,25 +57,31 @@ $deviceTemperature : [foo]`;
 
     it('Parser throws when a composite entity has a pattern.any entity as a child', function(done){
         let luFileContent = `$deviceTemperature : [p1; child2]
-# testIntent
-- I'm {p1}`;
+                            # testIntent
```
christopheranderson: why add whitespace here?

packages/lu/test/parser/lufile/parseFileContents.qnafile.test.js
```diff
@@ -32,11 +32,14 @@ describe('With parse file function', function() {
...
-                assert.equal(res.files[0].fileUri, 'https://download.microsoft.com/download/2/9/B/29B20383-302C-4517-A006-B0186F04BE28/surface-pro-4-user-guide-EN.pdf');
-            })
-            .catch(err => {throw(err)});
+            try {
+                let qna = new qnamaker()
+                let qnaList = [Blob1.qnaJsonStructure, Blob2.qnaJsonStructure]
+                qna.collate(qnaList)
+                assert.equal(qna.files[0].fileUri, 'https://download.microsoft.com/download/2/9/B/29B20383-302C-4517-A006-B0186F04BE28/surface-pro-4-user-guide-EN.pdf');
+            } catch(err){
+
```
christopheranderson: you dropped a `throw err` here

PR #299, opened 2019-10-30
several changes

ninggao:
(1) remove unused old files;
(2) change incorrect file names;
(3) add string ask templates;
(4) remove meta properties from schemas;
(5) fix an error in calculating optional properties
(6) fix help property bug


packages/dialog/test/commands/dialog/forms/activity.form.dialog
```diff
@@ -16,30 +16,7 @@
...
-            "type": "string",
-            "$mappings": ["Confirmation"],
-            "$templates":[]
-        },
-        "PropertyToChange": {
-            "type": "string",
-            "$mappings":["PROPERTYNAME"], 
-            "$templates":[]
-        }     
+        }
```
chrimc62: Don't you still need the $required to pull in the standard properties for all of the CDS schemas?

PR #319, opened 2019-10-30
removing ignored files

axelsrz:


packages/lu/test/parser/luisgen/.gitignore
```diff
@@ -0,0 +1,353 @@
+## Ignore Visual Studio temporary files, build results, and
```
munozemilio: Is this intended to be here?

cleemullins:
@axelsrz Can you sync with @munozemilio and resolve this and close today?

PR #325, opened 2019-11-01
[Spec] Lubuild -> bf luis:make

vishwacsena:
Adding spec for bringing forward the current [lubuild](https://github.com/microsoft/botbuilder-tools/tree/V.Future/packages/lubuild) functionality into BF-CLI.


specs/LUIS-MAKE.md
```diff
@@ -0,0 +1,145 @@
+# Lubuild -> bf luis:make
```
scheyal: Review rendered view. Formatting (especially for bulleted items) is broken.

specs/LUIS-MAKE.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Functional requirements
+1. Given an .lu files that describes one luis application and a [luis-configuration](#LUIS-configuration),
+    a. Find a matching LUIS application. 
+    b. If found uploads new model definition (as defined in .lu file) as new version.
+    c. Else Create a new LUIS application with the model definition (as defined in .lu file). This is the functional equivalent of `import application` command.
+    d Train and publish the new application.
+2. Given a collection of .lu files that describe one luis application across one or more supported [luis locales][2] and a [luis-configuration](#LUIS-configuration),
+    a. For each locale, do 1.a - 1.d
+3. LUIS supports up to 5 concurrent authoring API calls and up to 5 calls per second to the service per authoring key. The library should leverage this capability as a performance optimization and parallelize calls up to 5 applications (each .lu content is an application). Note that for each .lu content, we might need to do either create/ import followed by train and then publish. We cannot parallelize calls within that workflow but can parallelize across applications.
```
scheyal: What is the failure path? 2 failed, 3 succeeded? What is the state and corresponding messaging?

specs/LUIS-MAKE.md
```diff
@@ -0,0 +1,145 @@
...
+4. To determine if an application exists, we should use the following order of precendence - 
+    a. If the .lu file has a @app.name property, we should use that as the app name to find.
+    b. <BotName>(<environment>)-<luFileNameOrId>.<locale>.lu
```
scheyal: need \\ escape. rendering broken

scheyal: We mix here explicit/implicit detection. Not clear what is the application in (b). Is there a reason we shouldn't use only (a) - explicit form. Is the use case for (b) common? If not common practice or needed for other reasons I'd consider sticking with (a). Just a suggestion to consider.

specs/LUIS-MAKE.md
```diff
@@ -0,0 +1,145 @@
...
+5. To determine if a new version needs to be created, we will
```
scheyal: And if we determine yes or no, what is the consequent action? do we create and in which step?

specs/LUIS-MAKE.md
```diff
@@ -0,0 +1,145 @@
...
+    a. Get the latest published application version (to either PRODCUTION or STAGING) via call to [Get application info][3]
+    b. export the version via call to [Export application version][4]
+    c. [Transform][5] and perform a diff against the current LU JSON .vs. what's in latest version. Create a new version if the content is different or no content exists.
+6. If [luis-configuration](#LUIS-configuration) requests .dialog files be generated, then generate .dialog files with the newly created/ updated LUIS application information. See [dialog-files](#dialog-files) to learn more about the logic for creating .dialog and related files.
```
scheyal: dialog-files link/rendering is broken

specs/LUIS-MAKE.md
```diff
@@ -0,0 +1,145 @@
...
+    b. <BotName>(<environment>)-<luFileNameOrId>.<locale>.lu
+5. To determine if a new version needs to be created, we will
+    a. Get the latest published application version (to either PRODCUTION or STAGING) via call to [Get application info][3]
+    b. export the version via call to [Export application version][4]
+    c. [Transform][5] and perform a diff against the current LU JSON .vs. what's in latest version. Create a new version if the content is different or no content exists.
+6. If [luis-configuration](#LUIS-configuration) requests .dialog files be generated, then generate .dialog files with the newly created/ updated LUIS application information. See [dialog-files](#dialog-files) to learn more about the logic for creating .dialog and related files. 
+
+## Locale naming conventions
+When the CLI route is atttempting to infer the locale from the file name, we should expect the file names to be in this convention - 
+<fileName>.<locale>.lu
```
scheyal: \\ escape for rendering

specs/LUIS-MAKE.md
```diff
@@ -0,0 +1,145 @@
...
+
+<locale> must be one of the supported [luis locales][2].
+
+## Version naming conventions
+When creating a new application, version the application as `1`
+When creating a new version, increment the application version by 1. So versions would be named as 1, 2, 3, 4, ...
+If the prior version is not a number, then add -1, -2, -3 ... as a **suffix**. E.g. Initial version - `myversion` -> `myversion-1` -> `myversion-2` -> ...
```
scheyal: Ok. Just noting that we should follow common practices for versions if there are any (e.g. dot or dash separated).

specs/LUIS-MAKE.md
```diff
@@ -0,0 +1,145 @@
...
+
+## LUIS configuration
```
scheyal: Recommendations:
* For LUIS we store only a few API config values. As a principle, we should refrain from "bloating", doing too much in config as I think it can lead to user confusion. Keep only values that are likely to be static for application (such as auth key, region). See more [here](https://github.com/microsoft/botframework-cli/blob/master/specs/LUIS-CLI-Usage-Spec.md#luis-configuration-values)
* Filepaths (in/out): use common guidelines for CLI (use CWD unless path is specified and resolved on command line). 
  * Library gets the provided (resolved) path as is. If using STDIN, library gets buffer (or library only gets buffer?)
* LUIS (Ocp?) Cognitive Subscription key is already stored in CLI lib.
* Region is already stored in config. 
In summary, I suggest all these values should come from command line (other than what's already specified per link above).

specs/LUIS-MAKE.md
```diff
@@ -0,0 +1,145 @@
...
+6. [Optional] Output folder name to write out .dialog files generated. Defaults to `cwd()`
+7. [Optional] --dialog field to request that .dilog files be written out. Defaults to `false`
+
+## CLI Command
+
+`bf luis:make`
+
+| Parameters        | Description                                                                                                             |
+|-------------------|-------------------------------------------------------------------------------------------------------------------------|
+| --in              | lu file or folder                                                                                                       |
```
scheyal: Specify mandatory/optional and default for all values (e.g. STDIN).
Also, see cross-reference to [LUIS API Usage spec](https://github.com/microsoft/botframework-cli/blob/master/specs/LUIS-CLI-Usage-Spec.md) and [Contributing guidelines](https://github.com/microsoft/botframework-cli/blob/master/CONTRIBUTING.md) for respecting common flags & I/O behavior.

specs/LUIS-MAKE.md
```diff
@@ -0,0 +1,145 @@
...
+| --force           | force write                                                                                                             |
+| --authoringkey    | LUIS authoring key                                                                                                      |
```
scheyal: Sadly, we're calling it subscriptionKey to align with LUIS, QnAMaker, cognitive services terminology. We'd need to ensure people know it is *not* Azure subscription key...

specs/LUIS-MAKE.md
```diff
@@ -0,0 +1,145 @@
...
+| --botname         | bot name                                                                                                                |
+| --out             | [Optional] output location                                                                                              |
+| --culture         | [Optional] culture code for the content. Infer from .lu if available. Defaults to `en-us`                               |
+| --authoringregion | [Optional] LUIS authoring region                                                                                        |
```
scheyal: Follow LUIS API spec. "region"

specs/LUIS-MAKE.md
```diff
@@ -0,0 +1,145 @@
...
+| --environmentname | [Optional] nvironment name to include in LUIS app name                                                                  |
+| --dialog          | [Optional] write out .dialog files                                                                                      |
+| --fallbacklocale  | [Optional] locale to be used at the fall back if no locale specific recognizer is found. Only valid if --dialog is set. |
```
scheyal: Why not just locale? Explain in synopsis that is is fall back if...

specs/LUIS-MAKE.md
```diff
@@ -0,0 +1,145 @@
...
+| --in              | lu file or folder                                                                                                       |
+| --force           | force write                                                                                                             |
+| --authoringkey    | LUIS authoring key                                                                                                      |
+| --botname         | bot name                                                                                                                |
+| --out             | [Optional] output location                                                                                              |
+| --culture         | [Optional] culture code for the content. Infer from .lu if available. Defaults to `en-us`                               |
+| --authoringregion | [Optional] LUIS authoring region                                                                                        |
+| --environmentname | [Optional] nvironment name to include in LUIS app name                                                                  |
```
scheyal: Can there be multiple environments? According to LUIS API it is staging or prod. I suggested --staging true/false flag accordingly. We should align either way.

specs/LUIS-MAKE.md
```diff
@@ -0,0 +1,145 @@
...
+## CLI Command
+
+`bf luis:make`
+
+| Parameters        | Description                                                                                                             |
+|-------------------|-------------------------------------------------------------------------------------------------------------------------|
+| --in              | lu file or folder                                                                                                       |
+| --force           | force write                                                                                                             |
+| --authoringkey    | LUIS authoring key                                                                                                      |
+| --botname         | bot name                                                                                                                |
```
scheyal: In LUIS we refer to application by *name*. So it is the name of the application, "bot agonistic". Is this essentially the LUIS application name? If so, we should just refer to it as *name*.

specs/LUIS-MAKE.md
```diff
@@ -0,0 +1,145 @@
...
+| --out             | [Optional] output location                                                                                              |
+| --culture         | [Optional] culture code for the content. Infer from .lu if available. Defaults to `en-us`                               |
+| --authoringregion | [Optional] LUIS authoring region                                                                                        |
+| --environmentname | [Optional] nvironment name to include in LUIS app name                                                                  |
+| --dialog          | [Optional] write out .dialog files                                                                                      |
+| --fallbacklocale  | [Optional] locale to be used at the fall back if no locale specific recognizer is found. Only valid if --dialog is set. |
+
+
+## Writing out .dialog files
+All up:luis.settings.<environmentname>.<region>.json
```
scheyal: Explain: is this another config file? We are trying to have a single config file. Or is it a data file? Let's chat.

specs/LUIS-MAKE.md
```diff
@@ -0,0 +1,145 @@
...
+{
+    "$type": "Microsoft.MultiLanguageRecognizer",
+    "recognizers": {
+        "<locale>": "<lufilename>.<locale>.lu",
+        "": "<lufilename>.<fallbacklocale>.lu"
+    }
+}
+```
+
+## Interfaces
```
scheyal: @christopheranderson - FYI regarding lib export quality gates. Do we need another lib review for public consumption?

christopheranderson: If we're intending to make these APIs publicly consumable, I'd advocate we do a review, though we can do it here in this spec. That said, I see a number of problems in this interface. Should I leave my feedback here?

specs/LUIS-MAKE.md
vishwacsena: Fixed.

specs/LUIS-MAKE.md
vishwacsena: Clarified

specs/LUIS-MAKE.md
vishwacsena: Fixed

vishwacsena: clarified.

specs/LUIS-MAKE.md
vishwacsena: clarified.

specs/LUIS-MAKE.md
vishwacsena: fixed.

specs/LUIS-MAKE.md
vishwacsena: Clarified and fixed.

specs/LUIS-MAKE.md
scheyal: Use suffix instead of environment.

specs/LUIS-MAKE.md
scheyal: Also, break the dialog into its own command for CI/CD agility.
Keep fallbackLocale

specs/LUIS-MAKE.md
scheyal: Ok.

scheyal: File a bug to make this a single file

specs/LUIS-MAKE.md
scheyal: different discussion.

specs/LUIS-MAKE.md
```diff
@@ -0,0 +1,149 @@
...
+    - For each locale, do 1.a - 1.d
+3. LUIS supports up to 5 concurrent authoring API calls and up to 5 calls per second to the service per authoring key. The library should leverage this capability as a performance optimization and parallelize calls up to 5 applications (each .lu content is an application). Note that for each .lu content, we might need to do either create/ import followed by train and then publish. We cannot parallelize calls within that workflow but can parallelize across applications.
+    - Failures on any calls to LUIS should automatically initiate a wait-retry loop for up to 5 times. Repeat failures should be reported back with the actual error message from the service and the corresponding LU content (file name/ content itself).
+4. To determine if an application exists, we should use the following order of precedence. This is applicable for #1.a-c above to determine if an application exists already or we should create a new application - 
+    - (a) If the .lu file has a @app.name property, we should use that as the app name to find. **.OR.**
```
boydc2014: You mean literally using "@app.name = foo" or syntax like 
```
# ! app.name = foo
```
like we did in sectioned-lu

specs/LUIS-MAKE.md
```diff
@@ -0,0 +1,149 @@
...
+- When creating a new version, use the following order of precedence - 
+    - use @app.versionId if present in the LU content.
+    - increment the application minor version by 1. So versions would be named as 0.1, 0.2, 0.3, 0.4, ...
+    - If the prior version is not a number, then add -0.1, -0.2, -0.3 ... as a **suffix**. E.g. Initial version - `myversion` -> `myversion-0.1` -> `myversion-0.2` -> ...
+
+## LUIS configuration
+1. Location of the .lu file (absolute or relative path or file content via stdin for CLI)
+2. LUIS authoring key to use
+3. Bot name (text property that can be set to e.g. `WeatherBot`, `HRBot` etc). 
+4. [Optional] Default language - identifies .lu file's base language when a lang code is not explicitly included.
```
boydc2014: just to be 100% sure, language here means locale right? sometimes it's not exactly the same thing

specs/LUIS-MAKE.md
```diff
@@ -0,0 +1,149 @@
...
+7. [Optional] --dialog field to request that .dilog files be written out. Defaults to `false`
+
+## CLI Command
+
+`bf luis:up`
+
+| Parameters        | Description                                                                                                             |
+|-------------------|-------------------------------------------------------------------------------------------------------------------------|
+| --in              | lu file or folder                                                                                                       |
+| --force           | force write                                                                                                             |
```
boydc2014: force means?

specs/LUIS-MAKE.md
```diff
@@ -0,0 +1,149 @@
...
+
+```ts
+export class LuBuild {
+    public luConfig:LUISConfig;
+
+    public static async CreateOrUpdateApplication(luConfig:LUISConfig, delegate:(ID:string) : Content) : Promise<List<LUISApps>> {
+        return new Array<LuBuildResult>();
+    }
+
+    public async GenerateDeclarativeAssets(LuisAppList:List<LUISApps>):Promise<DialogFileContent>;
```
vishwacsena: Promise<Array<Content>>

specs/LUIS-MAKE.md
```diff
@@ -0,0 +1,149 @@
...
+
+export class LUISConfig {
+    public AuthoringKey:string;
+    public BotName:string;
+    public Culture:string = 'en-us';
+    public AuthoringRegion:string = 'westus';
+    public EnvironmentName:string = 'dev';
+    public GenerateDialogFileContent:boolean = false;
+    public FallBackLocale:string = 'en-us';
+    public IDs:List<string>;
```
vishwacsena: public LuContent:Array<Content> instead of IDs and Content

specs/LUIS-MAKE.md
vishwacsena: This is the authoring key and not the subscription key.

PR #328, opened 2019-11-04
Command line usage & form for bf:luis command group

scheyal:


scheyal:
> For CLI args, I thought we were using all lower case? I see some are camel case and others are not.

We will use camelCase. QnAMaker CLI is already using camel case. All lower case is not readable. It would have been best if we were case insensitive (and I hope we get there).

PR #333, opened 2019-11-06
Chatdown refactor for 4.7

jspruance:
This is a post-release refactor of the chatdown command code containing some cleanup and restructuring work that we weren't able to get to before the 4.6 release.

- Restructure command 'run' function so that it just contains flag parsing logic, so that it's easier to visually understand the flow

- move other logic (including file parsing / writing and transcript conversion logic into single-purpose methods on the chatdown command class

- Update unit tests to ensure backwards compatibility


jspruance:
I believe the Coveralls failure report is incorrect - I actually increased coverage with this change.

packages/chatdown/src/commands/chatdown/convert.ts
```diff
@@ -43,33 +44,27 @@ export default class ChatdownConvert extends Command {
...
+        throw new CLIError('No chat files found at: ' + flags.in)
       }
 
+      // step 2: convert chat data input to transcript(s)
+      const transcriptObjects = await this.convertToTranscriptObject(inputEntities, inputIsDirectory, flags.in)
+
+      // step 3: write output transcript data
+      await this.writeTranscripts(transcriptObjects, outputDirectory, flags.forced)
+
     } catch (err) {
```
munozemilio: Can we reduce the scope of this try catch to be only at the expected exception?

packages/chatdown/test/commands/chatdown/convert.test.ts
```diff
@@ -60,15 +66,22 @@ describe('chatdown:convert', () => {
     });
 
   it('should process all files when a glob is passed in with the -i argument, and the -o is passed in for the output directory', done => {
-    cp.exec(`node ./bin/run chatdown:convert -i "./test/utils/*.sample.chat" -o ./testout`, (error, stdout, stderr) => {
+    cp.exec(`node ./bin/run chatdown:convert -i "**/test/utils/*.sample.chat" -o ./testout`, (error, stdout, stderr) => {
```
munozemilio: Rather than changing the test, we should add a the new one and keep the old one too

packages/chatdown/src/commands/chatdown/convert.ts
```diff
@@ -34,7 +33,9 @@ export default class ChatdownConvert extends Command {
     try {
       const {flags} = this.parse(ChatdownConvert)
 
-      let inputIsDirectory = flags.in ? (flags.in.includes('*') || this.isDir(flags.in)) : false
+      let inputEntities: string[] | undefined = []
```
munozemilio: We should add tests to cover the following scenarios:
folder as input and file name as output
file name as input and folder name as output,
folder as input and folder as output where
Validation of the numbered file names when file already exists

packages/chatdown/src/commands/chatdown/convert.ts
jspruance: done

packages/chatdown/test/commands/chatdown/convert.test.ts
jspruance: agree, but what I found was that the previous test was actually invalid

packages/chatdown/src/commands/chatdown/convert.ts
jspruance: Tests added for:
file name as input and folder name as output,
folder as input and folder as output where
Validation of the numbered file names when file already exists

Thought we decided not to support filename as output (it does not currently), but let me know if you want me to add:
folder as input and file name as output

munozemilio: We should add those test to check the error is handled properly in case the user passes a filename as an output

jspruance: test case added

cleemullins:
Closing as no longer relevant.

Issue #300, opened 2019-10-30
Create Luis Plugin


munozemilio:
Create the new Luis plugin with the name: @microsoft/bf-luis-cli
Install the following dependencies:
```
    "azure-cognitiveservices-luis-authoring": "2.1.1",
    "ms-rest": "2.5.3"
```

jspruance:
This is complete - code merged into master branch.

PR #336, opened 2019-11-06
bf-luis-cli plugin initial commit

jspruance:
- initial plugin setup
- add initial dependencies

for issue:
https://github.com/microsoft/botframework-cli/issues/300


jspruance:
Note: I'm leaving in the boilerplate hello world command and tests for now - will replace with luis commands once I begin that work

PR #337, opened 2019-11-07
Create bf luis plugin commands

jspruance:


packages/luis/README.md
```diff
@@ -28,25 +28,24 @@ USAGE
 <!-- usagestop -->
 # Commands
 <!-- commands -->
-* [`oclif-example hello [FILE]`](#oclif-example-hello-file)
+* [`oclif-example luis:version:clone [FILE]`](#oclif-example-luisversionclone-file)
```
munozemilio: please modify the oclif bin entry in package.json to be 'bf' and the generate the doc

packages/luis/src/commands/luis/version/clone.ts
```diff
@@ -0,0 +1,62 @@
+/*!
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import {CLIError, Command} from '@microsoft/bf-cli-command'
+import {flags} from '@oclif/command'
```
munozemilio: please import flags from @microsoft/bf-cli-command

packages/luis/utils/index.js
```diff
@@ -0,0 +1,67 @@
+
```
munozemilio: can we move this into src/ and write it in ts?

packages/luis/utils/index.js
```diff
@@ -0,0 +1,67 @@
...
+const windowsHomedriveHome = () => process.env.HOMEDRIVE && process.env.HOMEPATH && path.join(process.env.HOMEDRIVE, process.env.HOMEPATH)
+const windowsUserprofileHome = () => process.env.USERPROFILE
+const windowsHome = () => windowsHomedriveHome() || windowsUserprofileHome()
+
+const packageName = pjson.name
+const platform = os.platform()
+const isWindows = platform === 'win32'
+const home = process.env.HOME || (isWindows && windowsHome()) || os.homedir() || os.tmpdir()
+
+const getConfigDir = (isWindows, home, pname) => {
```
munozemilio: No need to detect the OS. The command can pass this.config.configDir, which is the path to the config file. Qnamaker commands have an example of this.

packages/luis/README.md
jspruance: done

packages/luis/src/commands/luis/version/clone.ts
jspruance: done

packages/luis/utils/index.js
jspruance: done

packages/luis/utils/index.js
jspruance: yes, much better. i forgot we were doing this manually only outside of the context of a command before (post install script etc.). done.

packages/luis/README.md
```diff
@@ -12,10 +12,25 @@
...
+$ bf COMMAND
+running command...
+$ bf (-v|--version|version)
+@microsoft/bf-luis-cli/0.0.0 win32-x64 node-v10.16.3
+$ bf --help [COMMAND]
+USAGE
+  $ bf COMMAND
+...
+```
+<!-- usagestop -->
```
munozemilio: Please delete the usage

packages/luis/README.md
munozemilio: Link still appears as oclif-example

packages/luis/README.md
jspruance: usage deleted

packages/luis/README.md
jspruance: i'm relying on oclif too much to get it right! ok, this is corrected

packages/luis/src/commands/luis/version/clone.ts
```diff
@@ -0,0 +1,62 @@
...
+      }
+
+      try {
+        const latestVersion = await client.versions.clone(appId, versionId, options)
+        this.log(`App successfully cloned. Latest version is now: ${latestVersion}`)
+      } catch (err) {
+        throw new CLIError(`Failed to clone app: ${err}`)
+      }
+
+    } catch (err) {
```
munozemilio: Please remove this try catch since this is not applicable

jspruance: Do you mean the try / catch surrounding the 'throw new CLIError(`Failed to clone app: ${err}`)' message? I added that because I saw a requirement to have an error message specific to the given command. If you don't think we need it anyways, I can remove it.

munozemilio: The one with catch in line 54

jspruance: Outer try / catch removed

PR #350, opened 2019-11-12
Nigao/newtemplate

ninggao:


packages/dialog/templates/stringSetdatetimeV2.dialog.lg
```diff
@@ -11,7 +11,7 @@
     "actions": [
         {
             "$type": "Microsoft.IfCondition",
-            "condition": "contains(turn.schema.properties['@{property}'].$mappings,'datetimeV2'",
+            "condition": "contains(turn.schema.properties['@{property}'].$mappings,'datetimeV2')",
             "actions":[
```
chrimc62: This is not the right way to do this--you cannot rely on $mappings to contain the entity.  The right way is to use the "entity" property and set it to "datetimeV2".

packages/dialog/templates/stringSetdimension.dialog.lg
```diff
@@ -6,9 +6,14 @@
 - ```
 {
     "$schema": "@{appSchema}",
-    "$type": "Microsoft.SetProperty",
-    "property": "$@{property}",
-    "value": "@dimension"
-               
+    "$type": "Microsoft.OnSetProperty",
+    "property": "@{property}",
+    "actions": [
```
chrimc62: Missing entity.

PR #351, opened 2019-11-13
Adding luis:application:create cmd and tests / refactor utils and clo…

jspruance:
- create 'luis:application:create' cmd and tests
- refactor of utils to make input processing more generic
- update clone cmd to support utils refactor

### DIFF
--- a/packages/luis/src/commands/luis/application/create.ts
+++ b/packages/luis/src/commands/luis/application/create.ts
@@ -1,0 +1,62 @@
+/*!
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import {CLIError, Command, flags} from '@microsoft/bf-cli-command'
+
+const utils = require('../../../utils/index')
+
+export default class LuisApplicationCreate extends Command {
+  static description = 'Creates a new LUIS application'
+
+  static examples = [`
+    $ bf luis:version:clone --endpoint {ENDPOINT} --subscriptionKey {SUBSCRIPTION_KEY} --name {NAME} --culture {CULTURE}
+    --domain {DOMAIN} --description {DESCRIPTION} --initialVersionId {INITIAL_VERSION_ID} --usageScenario {USAGE_SCENARIO}
+  `]
+
+  static flags = {
+    help: flags.help({char: 'h'}),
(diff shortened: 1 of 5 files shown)


packages/luis/src/commands/luis/application/create.ts
```diff
@@ -0,0 +1,62 @@
...
+    description: flags.string({description: 'LUIS application description'}),
+    initialVersionId: flags.string({description: 'LUIS application initial version Id'}),
+    usageScenario: flags.string({description: 'LUIS application usage scenario'}),
+  }
+
+  async run() {
+    const {flags} = this.parse(LuisApplicationCreate)
+    const flagLabels = Object.keys(LuisApplicationCreate.flags)
+    const configDir = this.config.configDir
+    const configPrefix = 'luis__'
```
munozemilio: This should be moved to utils since it is the same for all commands

packages/luis/src/utils/index.ts
```diff
@@ -43,16 +43,16 @@ const getPropFromConfig = async (prop: string, configDir: string) => {
...
   let config = await getUserConfig(configDir)
   config = config ? filterConfig(config, prefix) : config
-  const input = {
-    appId: flags.appId || (config ? config.luis__appId : null),
-    endpoint: flags.endpoint || (config ? config.luis__endpoint : null),
-    subscriptionKey: flags.subscriptionKey || (config ? config.luis__subscriptionKey : null),
-    versionId: flags.versionId || (config ? config.luis__versionId : null),
-    targetVersionId: flags.targetVersionId || (config ? config.luis__targetVersionId : null)
-  }
+  const input: any = {}
```
munozemilio: IMO the previous implementation is clearer to understand what the programs is retrieving from the config, plus it is easier to control the flow

packages/luis/src/commands/luis/application/create.ts
jspruance: done

packages/luis/src/utils/index.ts
jspruance: But the previous implementation was hardcoded to one cmd (clone) so it wasn't flexible for general use. That's why I changed it to set the flags dynamically so that it works across the board.

PR #353, opened 2019-11-14
Initial check-in of dialog generation.

chrimc62:
This defines dialog:generate and is marked as preview.


packages/cli/package.json
```diff
@@ -10,6 +10,7 @@
   "dependencies": {
     "@microsoft/bf-chatdown": "1.0.0",
     "@microsoft/bf-cli-config": "1.0.0",
+    "@microsoft/bf-dialog": "1.0.0",
```
munozemilio: Is this going to be part of the 4.7 release?

packages/dialog/src/commands/dialog/generate.ts
```diff
@@ -0,0 +1,60 @@
...
+                (type, msg) => {
+                    if (type === gen.FeedbackType.message
+                        || type === gen.FeedbackType.error
+                        || (type === gen.FeedbackType.info && flags.verbose)) {
+                        this.progress(msg)
+                    }
+                })
+            return true;
+        } catch (e) {
+            this.thrownError(e)
```
munozemilio: You can throw CLIError imported from @microsoft/bf-cli-command and have the same behavior

packages/dialog/src/commands/dialog/generate.ts
```diff
@@ -0,0 +1,60 @@
...
+            let outDir = flags.output
+            if (!outDir) {
+                outDir = ppath.join(formName + '-resources')
+            }
+            await gen.generate(args.schema, outDir, flags.schema, flags.locale, flags.templates, flags.force,
+                (type, msg) => {
+                    if (type === gen.FeedbackType.message
+                        || type === gen.FeedbackType.error
+                        || (type === gen.FeedbackType.info && flags.verbose)) {
+                        this.progress(msg)
```
munozemilio: Can we use this.error directly?

packages/cli/package.json
chrimc62: As I understand the plan is to mark the dialog stuff as [Preview], but release as part of 4.7.

packages/dialog/src/commands/dialog/generate.ts
chrimc62: No.  The errors come from lots of different sources outside this object.

PR #354, opened 2019-11-14
support section and cross-train in bf lu

feich-ms:
merge section CRUD and cross-train functionalities into master


packages/lu/README.md
```diff
@@ -24,6 +25,7 @@
 * [`bf luis:generate:cs`](#bf-luisgeneratecs)
 * [`bf luis:generate:ts`](#bf-luisgeneratets)
 * [`bf luis:translate`](#bf-luistranslate)
+* [`bf luis:cross-train`](#bf-luiscross-train)
```
munozemilio: @scheyal, @vishwacsena  should we change the name to avoid the hyphen? We dont have hyphens in any other command

vishwacsena:
Fei,

Can we update this to drop the CLI head and just keep API for now?

I'll test the functionality and report back.

feich-ms:
@vishwacsena I removed cross-train cli related docs and ts file. But the cross-train api parameters need lu objects parsed from lu files and triggerIntent->dialog mapping dict parsed from config file. These file content extraction was implemented in crossTrain.ts file which was removed just now. If you want to use the cross-train API, you can refer to crossTrain.ts file in previous commit. Thanks, Fei.

packages/lu/README.md
feich-ms: removed the doc here

packages/lu/src/commands/luis/cross-train.ts
```diff
@@ -0,0 +1,95 @@
+/*!
```
munozemilio: @vishwacsena are we going to expose this in 4.7?

feich-ms: No, will not expose it. The reason I keep it here is for review and better understanding. I will remove it if there are no other issues.

packages/lu/src/parser/lu/luCrossTrainer.js
```diff
@@ -0,0 +1,244 @@
...
+
+module.exports = {
+    /**
+     * Do cross training among lu files
+     * @param {luObject[]} luObjectArray the luObject list to be parsed
+     * @param {any} crossTrainConfig cross train json config
+     * @returns {Map<string, LUResource>} Map of file id and luResource
+     * @throws {exception} Throws on errors. exception object includes errCode and text
+     */
+    luCrossTrain: async function (luObjectArray, crossTrainConfig) {
```
munozemilio: Can we split this into smaller functions to improve readability?

packages/lu/src/parser/lu/luCrossTrainer.js
```diff
@@ -0,0 +1,244 @@
...
+                    for (const section of luResource.Sections) {
+                        if (section.SectionType === LUSectionTypes.SIMPLEINTENTSECTION 
+                            || section.SectionType === LUSectionTypes.NESTEDINTENTSECTION) {
+                            intents.push(section);
+                        }
+                    }
+
+                    const destLuFileToIntent = triggerRules[fileId];
+                    for (const destLuFile of Object.keys(destLuFileToIntent)) {
+                        if (fileIdsFromInput.includes(destLuFile)) {
```
munozemilio: When possible invert the condition to reduce nested code e.g
```
if (!fileIdsFromInput.includes(destLuFile)) continue
```

packages/lu/src/parser/lu/luCrossTrainer.js
```diff
@@ -0,0 +1,244 @@
...
+        // Parse resources
+        let rootResource = resources.filter(r => r.id === rootResourceId)[0];
+        rootResource.visited = true;
+        this.mergeRootInteruptionToLeaves(rootResource, idToResourceMap, intentName);
+
+        return Array.from(idToResourceMap.values());
+    },
+
+    mergeRootInteruptionToLeaves: function (rootResource, result, intentName) {
+        if (rootResource && rootResource.children && rootResource.children.length > 0) {
```
munozemilio: NIT. Same case on inverted conditions

packages/lu/src/parser/lu/luCrossTrainer.js
```diff
@@ -0,0 +1,244 @@
...
+                    if (line.trim().startsWith('-')) {
+                        newLines.push('- ' + line.trim().slice(1).trim());
+                    } else if (line.trim().startsWith('##')) {
+                        newLines.push('## ' + line.trim().slice(2).trim());
+                    } else if (line.trim().startsWith('#')) {
+                        newLines.push('# ' + line.trim().slice(1).trim());
+                    }
+                })
+
+                newFileContent = newLines.join('\r\n');
```
munozemilio: should we use const NEWLINE = require('os').EOL for new lines? I think this will behave better in cross platform

packages/lu/src/utils/filehelper.ts
```diff
@@ -163,6 +169,68 @@ export async function detectLuContent(stdin: string, input: string) {
...
+
+  filesToParse = helpers.findConfigFiles(input, recurse)
+
+  if (filesToParse.length === 0) {
+    throw new CLIError('Sorry, no .lu files found in the specified folder.')
+  }
+  return filesToParse
+}
+
+export async function getConfigObject(input: string, recurse = false) {
```
munozemilio: If no command will be exposed, do we need to add this?

feich-ms: It seems no need to add this. I will remove all changes related with cli.

packages/lu/src/parser/lu/luCrossTrainer.js
feich-ms: done. Split into another two functions.

packages/lu/src/parser/lu/luCrossTrainer.js
feich-ms: Done. Also check other code pieces and optimized.

packages/lu/src/parser/lu/luCrossTrainer.js
feich-ms: Done

packages/lu/src/parser/lu/luCrossTrainer.js
feich-ms: Optimized.

packages/lu/src/commands/luis/cross-train.ts
feich-ms: I removed all the cross-train cli related changes and keep the API only in the last commit.

packages/lu/src/utils/filehelper.ts
feich-ms: removed.

feich-ms:
@munozemilio thanks for all the constructive comments. I have resoved all of them. And I also removed all the cli related changes and keep pure cross train api only. @vishwacsena Pls help to verify and approve if there are no other issues. Thanks.

PR #355, opened 2019-11-14
Adding luis:version:delete cmd

### LINKED ISSUE #304: Create command: bf luis:version:delete
This command should expose this [request](https://westeurope.dev.cognitive.microsoft.com/docs/services/luis-programmatic-apis-v3-0-preview/operations/5890b47c39e2bb052c5b9c3f) exposed in the azure sdk [client](https://github.com/Azure/azure-sdk-for-node/tree/master/lib/services/luis)

e.g 
```
const msRest = require("ms-rest");
const LUISAuthoringClient = require("azure-cognitiveservices-luis-authoring");
const subscriptionKey = "xxxxxxxxxx";
const token = { 
    "inHeader": { 
        'Ocp-Apim-Subscription-Key': subscriptionKey 
    }
};
const creds = new msRest.ApiKeyCredentials(token);
const client = new LUISAuthoringClient(creds, 'https://westus.api.cognitive.microsoft.com');
const appId = 'xxxxxxxxxx'
const versionId = "xxx";

client.baseUri = 'https://westus.api.cognitive.microsoft.com/luis/authoring/v3.0-preview/';

const exportVersion = async function(){
    let result = await client.versions.exportMethod(appId, versionId)
    console.log(JSON.stringify(result, null, 2)); 
}

const listPhraseLists = async function(){
    let result = await client.features.listPhraseLists(appId, versionId)
    console.log(JSON.stringify(result, null, 2)); 
}

exportVersion()
listPhraseLists()

```

jspruance:
fixes #304

### DIFF
--- a/packages/luis/src/commands/luis/version/delete.ts
+++ b/packages/luis/src/commands/luis/version/delete.ts
@@ -1,0 +1,49 @@
+/*!
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import {CLIError, Command, flags} from '@microsoft/bf-cli-command'
+
+const utils = require('../../../utils/index')
+
+export default class LuisVersionDelete extends Command {
+  static description = 'Deletes a version of a LUIS application'
+
+  static examples = [`
+    $ bf luis:version:delete --appId {APP_ID} --versionId {VERSION_ID} --endpoint {ENDPOINT} --subscriptionKey {SUBSCRIPTION_KEY}
+  `]
+
+  static flags = {
+    help: flags.help({char: 'h'}),
+    appId: flags.string({description: 'LUIS application Id'}),
+    versionId: flags.string({description: 'LUIS application version Id'}),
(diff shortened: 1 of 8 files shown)


packages/luis/src/commands/luis/version/delete.ts
```diff
@@ -0,0 +1,49 @@
...
+      subscriptionKey,
+    } = await utils.processInputs(flags, flagLabels, configDir)
+
+    const requiredProps = {appId, versionId, endpoint, subscriptionKey}
+    utils.validateRequiredProps(requiredProps)
+
+    const client = utils.getLUISClient(subscriptionKey, endpoint)
+
+    try {
+      await client.versions.deleteMethod(appId, versionId)
```
jspruance: the azure client 'deleteMethod' returns null, so I just await it and output a success message if there's no error. their api is a little inconsistent.

munozemilio: @christopheranderson should we open a bug against the sdk on details like this?

christopheranderson: yeah, definitely should.

Issue #47, opened 2019-07-12
DCR: as a bot developer, I would like to be able to have a way of programmatically push and pull LUIS models from the LUIS portal


gabog:
The current luis CLI provides some primitives to create, publish, train, etc. LUIS models but these are too granular and require writing some additional powershell to wrap them into common operations. 

As a developer, I would like to simply say:

```
luis push -luFile FlightBooking.lu -luisAppId MyAppId-luisAuthoringKey myAuthoringKey -luisSubscriptionKey mySubscriptionKey -luisAppVersion myVersion
```
And have my model created, trained, published and ready to use. 
- luisSubscriptionKey: can be optional but it should get assigned to my app if provided
- luisAppVersion should be optional and If it is not provided, it should just overwrite the latest or create a 0.1 one.

To get a a model, I would like to just say:

```
luis pull-luFile FlightBooking.lu -luisAppId MyAppId-luisAuthoringKey myAuthoringKey -luisAppVersion myVersion
```
And get a lu file with the app from LUIS:
- luisAppVersion should be optional and If it is not provided, it should just get the currently active and published model.

## Additional info

- A sample LuisPull written in powershell can be found [here](https://github.com/microsoft/botbuilder-dotnet/blob/gabog/CoreBotTests/tests/Microsoft.Bot.Builder.TestBot/CognitiveModels/luisPull.ps1)
- A sample LuisPush written in powershell can be found [here](https://github.com/microsoft/botbuilder-dotnet/blob/gabog/CoreBotTests/tests/Microsoft.Bot.Builder.TestBot/CognitiveModels/luisPush.ps1)
- These two scripts rely on a helper powershell file that uses the LUIS primitives and can be found [here](https://github.com/microsoft/botbuilder-dotnet/blob/gabog/CoreBotTests/tests/Microsoft.Bot.Builder.TestBot/CognitiveModels/luis_functions.ps1)

scheyal:
BF LUIS is R7

munozemilio:
Luis API will be done in R7 any parsing to Lu can be added in R8

PR #356, opened 2019-11-15
Adding luis:version:export cmd

jspruance:


jspruance:
fixes #305

packages/luis/src/commands/luis/version/export.ts
```diff
@@ -0,0 +1,56 @@
...
+    const requiredProps = {appId, versionId, endpoint, subscriptionKey}
+    utils.validateRequiredProps(requiredProps)
+
+    const client = utils.getLUISClient(subscriptionKey, endpoint)
+
+    try {
+      const appJSON = await client.versions.exportMethod(appId, versionId)
+      if (appJSON) {
+        const outputPath = out ? out : ''
+        await utils.writeOutput(outputPath, appJSON, force)
```
munozemilio: Although it is very clean and helps to reduce code duplication having this on the utils file, I think it is easier to understand the command flow by keeping the flags processing in the command. e.g 
```
if (flags.out) {
 writeToFile()
} else {
writeToConsole()
}
```

packages/luis/src/utils/index.ts
```diff
@@ -58,7 +57,42 @@ const validateRequiredProps = (configObj: any) => {
...
+    stats = fs.statSync(path)
+  } catch {
+    return false
+  }
+  return stats.isDirectory()
+}
+
+const writeToConsole = (outputContents: string) => {
+  const output = JSON.stringify(outputContents, null, 2)
+  process.stdout.write('App successfully exported\n')
```
munozemilio: Please keep command responsibilities at the command level. e.g 
```
// Command class
writeToConsole()
this.log(App successfully exported)
```

packages/luis/src/utils/index.ts
```diff
@@ -58,7 +57,42 @@ const validateRequiredProps = (configObj: any) => {
...
+  process.stdout.write(output, 'utf-8')
+}
+
+const writeToFile = async (outputLocation: string, content: any, force: boolean) => {
+  const validatedPath = utils.validatePath(outputLocation, '', force)
+  await fs.ensureFile(outputLocation)
+  await fs.writeJson(validatedPath, content, {spaces: 2})
+  process.stdout.write(`File successfully written: ${validatedPath}`)
```
munozemilio: Please keep command responsibilities(logging) at the command level. e.g 
```
// Command class
const path = writeToFile()
this.log(`File successfully written: ${path}`)
```

jspruance: I actually wanted to do that too, but with this model, the tests have a hard time reading the message since 'ctx.stdout' typically only reads the first line. need to think about how to overcome that:

expect(ctx.stdout).to.contain('File successfully written')

munozemilio: This.log uses process.stdout.write under the hood

packages/luis/src/commands/luis/version/export.ts
jspruance: Moved flag parsing to cmd

packages/luis/src/utils/index.ts
jspruance: refactored

packages/luis/src/utils/index.ts
jspruance: refactored

packages/luis/src/utils/index.ts
```diff
@@ -36,17 +36,27 @@ const getLUISClient = (subscriptionKey: string, endpoint: string) => {
   return luisClient
 }
```
jspruance: I added this util function in order to limit the allowed config vars to 'appId, region, subscriptionKey, versionId' per Eyal's specs

munozemilio:
Fixes #47

PR #360, opened 2019-11-18
LUIS application create - update to bring in line w latest spec

jspruance:


jspruance:
This is a second pass to bring in line w latest specs:
https://github.com/microsoft/botframework-cli/blob/master/specs/LUIS-CLI-Usage-Spec.md

Still outstanding: 'tokenizerVersion' - not currently supported in azure client

munozemilio:
Fixes #47

PR #363, opened 2019-11-19
chore: Swap to rush [IGNORE]

christopheranderson:
This PR switches our build system to use rush and pnpm.

This is a WIP, will update notes will full change list before asking for review.


munozemilio:
Looks good to me

PR #367, opened 2019-11-20
Adding luis:application:import cmd

jspruance:


jspruance:
- added luis:application:import cmd
- updated bf 'command' to expose stdin reading function
- disregard changes to 'luis:application:create' here - this was a 'partial handcrafted artisnal merge' so that it would work with latest pending changes

packages/command/src/index.ts
```diff
@@ -5,5 +5,6 @@
 
 import {CLIError, Command, flags} from './command'
 import utils from './utils'
-export {CLIError, Command, flags, utils}
+import ReadPipedStdin from './readpipeddata'
```
munozemilio: This is no needed, we are exposing this functionality as part of the command class

packages/luis/src/utils/index.ts
```diff
@@ -18,6 +18,20 @@ const filterConfig = (config: any, prefix: string) => {
     }, {})
 }
 
+const getInputFromFileOrStdin = async (args: any): Promise<string> => {
```
munozemilio: I dont think we have to do this here

packages/luis/src/utils/index.ts
```diff
@@ -55,6 +69,10 @@ const processInputs = async (flags: any, flagLabels: string[], configDir: string
   flagLabels
     .filter(flag => flag !== 'help')
     .map((flag: string) => {
+      if (flag === 'in') {
```
munozemilio: Should we exclude this flag?

jspruance: meaning remove it from input processing in utils? i don't think so - i think we should stay consistent

packages/command/src/index.ts
jspruance: refactored. after looking at it though, i still think it should live in utils because we have all other read / write / parse functionality there, so now there is this one-off. maybe there's a deeper inconsistency with how we achieve code reuse - through class inheritance vs utility functions. maybe we should have done it all one way or another. anyways...this works for now.

packages/luis/src/utils/index.ts
jspruance: refactored

packages/luis/src/commands/luis/application/import.ts
```diff
@@ -0,0 +1,65 @@
...
+    try {
+      const newAppId = await client.apps.importMethod(JSON.parse(appJSON), undefined)
+      this.log(`App successfully imported with id ${newAppId}.`)
+    } catch (err) {
+      throw new CLIError(`Failed to import app: ${err}`)
+    }
+  }
+
+  async getImportJSON(input: string) {
+    if (input) {
```
munozemilio: Stdin should override flags.in

munozemilio: If no stdin or flags.in value passed, CLIError should be thrown asking the user to provide input

jspruance:
fixes #367

packages/luis/src/commands/luis/application/import.ts
jspruance: added error & unit test

packages/luis/src/utils/index.ts
jspruance: leaving as is

WTobor:
Hi! When this changes will be available on npm?

PR #371, opened 2019-11-20
Adding luis:application:list cmd

jspruance:


packages/luis/src/commands/luis/application/list.ts
```diff
@@ -0,0 +1,58 @@
...
+  static examples = [`
+    $ bf luis:application:list --endpoint {ENDPOINT} --subscriptionKey {SUBSCRIPTION_KEY} --take 3
+    $ bf luis:application:list --endpoint {ENDPOINT} --subscriptionKey {SUBSCRIPTION_KEY} --out {PATH_TO_JSON_FILE}
+  `]
+
+  static flags = {
+    help: flags.help({char: 'h'}),
+    endpoint: flags.string({description: 'LUIS endpoint hostname'}),
+    subscriptionKey: flags.string({description: 'LUIS cognitive services subscription key (aka Ocp-Apim-Subscription-Key)'}),
+    out: flags.string({char: 'o', description: 'Path to the directory where the exported file will be placed.'}),
```
munozemilio: Spec states: Output results to specified file in JSON format, otherwise prints to STDOUT (optional)
Which suggests a file name can also be set as output, in the other hand, for this command I don't think there is a lot of value printing this to a file. @scheyal please comment

jspruance:
fixes #308

packages/luis/src/commands/luis/application/list.ts
jspruance: suggest we just move fwd since it's already coded...discuss removing --out (if desired) later

munozemilio: Can we set the description as the spec please?

jspruance: updated

munozemilio: not matching yet

jspruance: sorry...check again now

PR #378, opened 2019-11-22
Adding luis:version:rename cmd

jspruance:


packages/luis/src/commands/luis/application/show.ts
```diff
@@ -0,0 +1,25 @@
+import {Command, flags} from '@oclif/command'
```
munozemilio: Is this intended to be merged?

jspruance: no - removed!

PR #383, opened 2019-11-22
Fix to BF-LU to exclude empty descriptors on intents (from JSON -> LU)

vishwacsena:
Fixing a bug where BF-LU library was incorrectly writing out a `@ intent intentName` even when the intent has an empty list of descriptors (usesFeature)


packages/lu/.vscode/launch.json
```diff
@@ -0,0 +1,32 @@
+{
```
munozemilio: I think we are adding all launch configs at the root

PR #387, opened 2019-11-23
chore: fix dialog debugging

christopheranderson:
## Goals

Fixes the dialog debugging experience and address sourcemap issues.

## Changes

- Move launch.json to project root & fix paths
- Change settings to be compatible with new mocha.opts
- Move static properties to bottom of class to get around this issue: https://github.com/microsoft/TypeScript/issues/35307

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -8,59 +8,67 @@
             "type": "node",
             "request": "launch",
             "name": "Dialog Merge Tests",
-            "program": "${workspaceFolder}/node_modules/mocha/bin/_mocha",
-            "cwd": "${workspaceFolder}",
+            "program": "${workspaceFolder}/packages/dialog/node_modules/mocha/bin/_mocha",
+            "cwd": "${workspaceFolder}/packages/dialog",
             "args": [
-                "-u",
-                "tdd",
                 "--timeout",
                 "999999",
                 "--colors",
-                "${workspaceFolder}/test/commands/dialog/merge.test.ts"
+                "-g",
+                ".*Test schema merge.*"
             ],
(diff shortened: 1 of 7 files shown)


.vscode/launch.json
```diff
@@ -77,14 +85,14 @@
             "type": "node",
             "request": "launch",
             "name": "Generate Sandwich Form",
-            "program": "${workspaceFolder}/bin/run",
-            "preLaunchTask": "tsc: build - tsconfig.json",
```
chrimc62: For this kind of thing it is definitely nice to be able to build before launching.  What is the right pattern for this?

chrimc62: Maybe ${defaultBuildTask}?

.vscode/launch.json
```diff
@@ -77,14 +85,14 @@
...
+            "program": "${workspaceFolder}/packages/dialog/bin/run",
+            "sourceMaps": true,
```
chrimc62: Find to be explicit, but isn't sourceMaps = true the default?  For that matter, they don't even list it as an option in the docs. https://code.visualstudio.com/Docs/editor/debugging

packages/dialog/src/commands/dialog/generate.ts
```diff
@@ -57,4 +38,24 @@ export default class GenerateDialog extends Command {
     progress(msg: string): void {
         this.error(msg)
     }
+
```
chrimc62: This is annoying especially given that style rules complain about doing this.  I know you have a bug out for typescript, but maybe we should disable the rule?  Or maybe I shouldn't have the tslint extension...

PR #394, opened 2019-11-27
Chrimc/form

chrimc62:
Integrate $kind change.
Add $public.
Add lg expansion to schema.


packages/dialog/src/commands/dialog/merge.ts
```diff
@@ -313,14 +313,14 @@ export default class DialogMerge extends Command {
     }
 
     processRoles(definitions: any, metaSchema: any): void {
-        for (let type in definitions) {
-            this.walkJSON(definitions[type], (val: any, _obj, key) => {
+        for (let kind in definitions) {
```
christopheranderson: [nit] For `for in` statements, it is best practice to use a `hasOwnProperty` filter.

```js
for (const prop in obj) {
  if (obj.hasOwnProperty(prop)) {
    console.log(`obj.${prop} = ${obj[prop]}`);
  } 
}
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in

I personally prefer `for of`. In this case I'd recommend:

```ts
for (const kind of Object.keys(definitions)) {
```

I'll be adding some tslint rules to enforce this, so you can ignore for now if you'd like.

packages/dialog/src/commands/dialog/merge.ts
```diff
@@ -329,50 +329,50 @@ export default class DialogMerge extends Command {
         }
     }
 
-    processRole(role: string, elt: any, type: string, definitions: any, metaSchema: any, key?: string): void {
-        const prefix = 'unionType('
+    processRole(role: string, elt: any, kind: string, definitions: any, metaSchema: any, key?: string): void {
```
christopheranderson: [nit] you should avoid any types if you do have an expected type

packages/dialog/src/commands/dialog/merge.ts
```diff
@@ -329,50 +329,50 @@ export default class DialogMerge extends Command {
...
+        const prefix = 'union('
         if (role === 'expression' || role === 'lg' || role === 'memoryPath') {
-            if (elt.type) {
+            if (elt.kind) {
                 this.error(`${this.currentFile}:error: $role ${role} must not have a type.`)
```
christopheranderson: This error message should be "kind" now?

christopheranderson: is this a terminal error? Should you exit the function?

packages/dialog/src/commands/dialog/merge.ts
```diff
@@ -42,7 +42,7 @@ export default class DialogMerge extends Command {
     async mergeSchemas(patterns: string[], output?: string, branch?: string, update?: boolean, verbose?: boolean): Promise<boolean> {
         this.verbose = verbose
         this.failed = false
-        this.missingTypes = new Set()
+        this.missingKinds = new Set()
```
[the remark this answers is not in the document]
chrimc62: Because this is an output of the action, i.e. it lists the errors for layers above to handle.  (In this case the CLI wrapper.)

packages/dialog/src/commands/dialog/merge.ts
chrimc62: Not sure I buy this.  We know the is a plain old JSON object, not a class.

packages/dialog/src/commands/dialog/merge.ts
chrimc62: The linting rules I have generally complain if you do not--maybe when you add consistent ones I won't see that.  I also think that if you are buying into typescript it is not unreasonable to be explicit.

packages/dialog/src/commands/dialog/merge.ts
chrimc62: Nice catch--fixed.

chrimc62: No.  The merge process tries to capture all of the errors that do not stop parsing.

PR #401, opened 2019-12-04
Adding luis:init cmd

jspruance:


packages/luis/src/utils/index.ts
```diff
@@ -34,6 +37,36 @@ const getUserConfig = async (configPath: string) => {
...
   }
+  return null
+}
+
+const createConfigFile = async (configPath: string) => {
+  await fs.mkdirp(configPath)
+  await fs.writeFile(path.join(configPath, 'config.json'), JSON.stringify({}, null, 2))
+}
+
+const promptSaveConfig = async (flags: any, configPath: string) => {
```
munozemilio: Since this is related to the init, should this be in the command file as well?

packages/luis/src/utils/index.ts
```diff
@@ -34,6 +37,36 @@ const getUserConfig = async (configPath: string) => {
...
+  let userConfig = await getUserConfig(configPath)
+  if (userConfig === null) {
+    await createConfigFile(configPath)
+    userConfig = {}
+  }
+  const saveConfigOptIn = await cli.prompt('Would you like to save the provided values to your global config file? (Y/N)')
```
munozemilio: cli.confirm provides the checking on assertion

packages/luis/src/utils/index.ts
```diff
@@ -34,6 +37,36 @@ const getUserConfig = async (configPath: string) => {
...
+    flagLabels.map(label => {
+      userConfig[`${configPrefix}${label}`] = flags[label]
+    })
+    await writeUserConfig(userConfig, configPath)
+    return true
+  }
+  return false
+}
+
+const writeUserConfig = async (userconfig: any, configPath: string) => {
```
munozemilio: QnAMaker provides the ability to override the config when a resource is created...should we call this when a new app is imported?

packages/luis/src/utils/index.ts
jspruance: done

packages/luis/src/utils/index.ts
jspruance: done

packages/luis/src/utils/index.ts
jspruance: added this option to application:create via 'save' flag

packages/luis/package.json
```diff
@@ -40,9 +40,11 @@
     "@microsoft/bf-cli-command": "1.0.0",
     "@oclif/command": "^1.5.19",
     "@oclif/config": "^1.13.3",
+    "@oclif/errors": "^1.2.2",
     "@types/sinon": "^7.5.0",
     "azure-cognitiveservices-luis-authoring": "2.1.1",
     "azure-cognitiveservices-luis-runtime": "~3.0.0",
```
munozemilio: lets pin the exact version

packages/luis/package.json
```diff
@@ -40,9 +40,11 @@
     "@microsoft/bf-cli-command": "1.0.0",
     "@oclif/command": "^1.5.19",
```
munozemilio: please use ~ for oclif dependencies

packages/luis/src/commands/luis/application/query.ts
```diff
@@ -11,16 +11,16 @@ export default class LuisApplicationQuery extends Command {
   static description = 'Queries application for intent predictions'
 
   static examples = [`
-    $ bf luis:application:query --endpoint {ENDPOINT} --subscriptionKey {SUBSCRIPTION_KEY} --appId {APP_ID} --query {QUERY} --versionId {INITIAL_VERSION_ID}
+    $ bf luis:application:query --endpoint {ENDPOINT} --subscriptionKey {SUBSCRIPTION_KEY} --appId {APP_ID} --query {QUERY} --slot {SLOT_NAME}
```
munozemilio: Since Slot is either production or staging is there a way to make this a boolean to indicate if it is prod or not? This will avoid typos introduced by the user

packages/luis/package.json
jspruance: updated

packages/luis/src/commands/luis/application/query.ts
jspruance: ok, working on it...

PR #405, opened 2019-12-04
build: fix install issues

christopheranderson:
## Changes

- Fix missing dependency in lu that wasn't found in testing
- Fix version step to generate manifest.json and remove commands path

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,12 +1,58 @@
 {
   "name": "@microsoft/botframework-cli",
-  "version": "1.0.0",
+  "version": "4.7.0-previewCHRIS",
   "description": "One-stop-shop CLI to manage your bot’s resources. BF CLI and AZ CLI together cover your end-to-end bot development workflow needs.",
   "author": "Microsoft",
   "bin": {
     "bf": "./bin/run"
   },
   "bugs": "https://github.com/microsoft/botframework-cli",
+  "engines": {
+    "node": ">=8.0.0"
+  },
+  "files": [
+    "/bin",
+    "/lib",
+    "/scripts",
+    "/npm-shrinkwrap.json",
+    "/oclif.manifest.json"
+  ],
+  "homepage": "https://github.com/microsoft/botframework-cli",
+  "keywords": [
+    "oclif"
+  ],
+  "license": "MIT",
+  "main": "lib/index.js",
+  "oclif": {
+    "bin": "bf",
+    "plugins": [
+      "@oclif/plugin-help",
+      "@microsoft/bf-chatdown",
+      "@microsoft/bf-cli-config",
+      "@microsoft/bf-dialog",
+      "@microsoft/bf-qnamaker",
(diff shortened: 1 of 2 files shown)


packages/cli/package.json
```diff
@@ -1,45 +1,12 @@
 {
   "name": "@microsoft/botframework-cli",
-  "version": "1.0.0",
+  "version": "4.7.0-previewCHRIS",
```
chrimc62: Auto-generated?

munozemilio: this looks a bit hardcoded :)

christopheranderson: fixed in #b328471

packages/cli/package.json
```diff
@@ -77,14 +43,48 @@
...
+  "types": "lib/index.d.ts",
+  "dependencies": {
+    "@microsoft/bf-chatdown": "1.0.0",
+    "@microsoft/bf-cli-config": "1.0.0",
+    "@microsoft/bf-dialog": "1.0.0",
+    "@microsoft/bf-lu": "1.0.0",
+    "@microsoft/bf-qnamaker": "1.0.0",
+    "@oclif/command": "~1.5.13",
+    "@oclif/config": "~1.13.0",
+    "@oclif/errors": "~1.2.2",
```
chrimc62: This is the only thing you added?  Any particular reason you moved where all of these were in the file--made it hard to see what the change was.

packages/lu/package.json
```diff
@@ -58,7 +58,8 @@
     "lodash": "^4.17.15",
     "node-fetch": "^2.1.2",
     "semver": "^5.5.1",
-    "tslib": "^1.10.0"
+    "tslib": "^1.10.0",
+    "uuid": "~3.3.3"
```
chrimc62: This was the missing dependency?

packages/cli/package.json
christopheranderson: fixed

packages/cli/package.json
christopheranderson: I added errors so that ./bin/run works when trying to debug this in the future from local.

They got moved around by rush's formatter when I ran `rush add -p @oclif/errors`. I agree it's unfortunately it touched so many files. If we feel strongly, I can revert and manually modify the package.json.

packages/lu/package.json
christopheranderson: This was breaking lu. Vishwac was hitting this locally as well.

christopheranderson: but the core of the issue was that we were pointing at a non-existent directory that previously wasn't being stat'd because we included a manifest file with an empty commands object. I removed the manifest file unintentionally when I removed the doc step from the build because it was hanging in CI while prompting for whether telemetry was ok to collect.

christopheranderson: So this wasn't the issue you were facing, but you might have eventually hit it if you were doing lots of lu stuff.

Issue #373, opened 2019-11-21
[bf qnamaker:kb:create] Can't create Knowledge Bases


Nerconer:
I get an error when I try to run the command:
`$ bf qnamaker:kb:create`

Error: Cannot find module './../../../../utils/payloads/createKb'
    at Module.patchedRequire [as require] (/usr/local/lib/node_modules/@microsoft/botframework-cli/node_modules/diagnostic-channel/dist/src/patchRequire.js:14:46)
    at Object.<anonymous> (/usr/local/lib/node_modules/@microsoft/botframework-cli/node_modules/@microsoft/bf-qnamaker/lib/commands/qnamaker/kb/create.js:12:22)

munozemilio:
Hello @Nerconer, thanks for bringing this up. Could you please provide the cli version and OS you are running on please?

Nerconer:
Hi, the version of the CLI I am using is 
`$ bf -v`
`@microsoft/botframework-cli/4.6.0 linux-x64 node-v8.10.0`
Regarding the OS, I am on Windows 10 Enterprise and Ubuntu's terminal downloaded from Microsoft Store to use the CLI. The current version of Ubuntu is:
`$ lsb_release -a`
`No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 18.04.2 LTS
Release:        18.04
Codename:       bionic`

munozemilio:
Thanks for the information @Nercorner, I will try to reproduce in the same environment and get back to you as soon as possible.

munozemilio:
Hello @Nerconer, just to keep you updated, for some reason one of our transient dependencies is not performing in Ubuntu. In the meantime I can suggest you to use cmd. We will keep working on this to make it work in the environment you specified

munozemilio:
There is an issue opened against App insights team [here](https://github.com/microsoft/ApplicationInsights-node.js/issues/573)

jspruance:
I've submitted a fix for this issue - PR here:
https://github.com/microsoft/botframework-cli/pull/413/files

munozemilio:
Hello @Nerconer, this is fixed in 4.7.0 which is already released.

Nerconer:
Works great, thank you!

PR #413, opened 2019-12-10
Fixing casing on createkb require

jspruance:


packages/qnamaker/src/commands/qnamaker/kb/create.ts
```diff
@@ -9,7 +9,7 @@ const qnaconfig = require('../../../../utils/qnaconfig')
 const qnamaker = require('./../../../../utils/index')
 const Knowledgebase = require('./../../../../utils/api/knowledgebase')
 const Endpointkeys = require('./../../../../utils/api/endpointkeys')
-const createKbJSON = require('./../../../../utils/payloads/createKb')
+const createKbJSON = require('./../../../../utils/payloads/createkb')
```
jspruance: correcting the casing here - linux is case sensitive, so this fails on Ubuntu (re issue #373)

PR #419, opened 2019-12-11
Formatting publish response json

jspruance:


jspruance:
Small change to format json response for publish cmd

jspruance:
- Format publish response JSON
- Remove success message from query response

PR #421, opened 2019-12-12
fix bugs of lu lexer

feich-ms:
1. fix lu lexer bug
2. remove #!/usr/bin/env node in three files to make them usable in composer client which is not node environment


vishwacsena:
@feich-ms can you take a look at the merge conflicts? 

@munozemilio for PR review

feich-ms:
Conflicts resoved. @munozemilio could you please approve?

PR #422, opened 2019-12-12
lubuild support in bf cli

feich-ms:
support new lubuild functionalities in bf cli lu package based on the spec here https://github.com/microsoft/botframework-cli/blob/master/specs/LUIS-MAKE.md


packages/lu/src/parser/lubuild/lubuildClasses.ts
```diff
@@ -0,0 +1,71 @@
+export class DialogFileContent {
```
boydc2014: types.ts?

packages/lu/src/parser/lubuild/lubuildCore.ts
```diff
@@ -0,0 +1,258 @@
+import {LUISConfig, Content, DialogFileContent} from './lubuildClasses'
```
boydc2014: core.ts or main.ts, or index.ts?

packages/lu/package.json
```diff
@@ -48,14 +48,17 @@
...
     "antlr4": "^4.7.2",
+    "await-delay": "^1.0.0",
     "chalk": "2.4.1",
     "console-stream": "^0.1.1",
     "deep-equal": "^1.0.1",
     "fs-extra": "^8.1.0",
     "get-stdin": "^6.0.0",
     "intercept-stdout": "^0.1.2",
     "lodash": "^4.17.15",
+    "luis-apis": "2.5.1",
```
boydc2014: this version of luis-apis not support directly publish .lu, right? we are still publishing after converting to json ?

packages/lu/src/parser/lubuild/lubuildCore.ts
feich-ms: core.ts is ok, but there is no main.ts and index.ts. The cli logic is in up.ts whose name is used as the command name: bf luis:up

packages/lu/package.json
feich-ms: yes, currently the luis json content is still used by this PR

boydc2014:
The spec PR is merged, so please update the link to the version in master

packages/lu/src/parser/lubuild/core.ts
```diff
@@ -0,0 +1,284 @@
...
+            }
+
+            currentApp.versionId = newVersionId;
+            recognizer.versionId = newVersionId;
+
+            const options: any = {};
+            options.versionId = newVersionId;
+            process.stdout.write(`${recognizer.getLuPath()} creating version=${newVersionId}\n`);
+            await client.versions.importMethod(<AzureRegions>config.AuthoringRegion, <AzureClouds>"com", recognizer.getAppId(), currentApp, options);
+            await delay(500);
```
boydc2014: why we have to delay 500?

packages/lu/src/parser/lubuild/core.ts
```diff
@@ -0,0 +1,284 @@
...
+            process.stdout.write(`${recognizer.getLuPath()} no changes\n`);
+            return false;
+        }
+    }
+
+    private static async publishApplication(config: LUISConfig, client: LuisAuthoring, recognizer: LuisRecognizer): Promise<void> {    
+        process.stdout.write(`${recognizer.getLuPath()} waiting for training for version=${recognizer.versionId}...`);
+        let done = true;
+        do {
+            await delay(5000);
```
boydc2014: delay 5000?

boydc2014:
Another question is do we have concurrency support now? 5 concurrent thread is allow in LUIS API

packages/lu/package.json
vishwacsena: We should stop using the old luis-apis and use the new library in BF CLI - @microsoft/bf-luis-cli

packages/lu/src/commands/luis/up.ts
```diff
@@ -0,0 +1,141 @@
...
+const path = require('path');
+const fs = require('fs-extra');
+const fileHelper = require('./../../utils/filehelper');
+const fileExtEnum = require('./../../parser/utils/helpers').FileExtTypeEnum
+
+export default class LuisUp extends Command {
+  static description = 'Build lu files and train and publish luis applications'
+
+  static examples = [`
+    $ bf luis:up --in {INPUT_FILE_OR_FOLDER} --authoringkey {AUTHORING_KEY} --botname {BOT_NAME} --dialog {true}
```
vishwacsena: The latest is to go with bf luis:build

packages/lu/src/commands/luis/up.ts
```diff
@@ -0,0 +1,141 @@
...
+  `]
+
+  static flags: any = {
+    help: flags.help({ char: 'h' }),
+    in: flags.string({ char: 'i', description: 'lu file or folder' }),
+    authoringkey: flags.string({ description: 'LUIS authoring key' }),
+    botname: flags.string({ description: 'bot name' }),
+    out: flags.string({ description: '[Optional] output location' }),
+    culture: flags.string({ description: '[Optional] culture code for the content. Infer from .lu if available. Defaults to en-us', default: 'en-us' }),
+    authoringregion: flags.string({ description: '[Optional] LUIS authoring region', default: 'westus' }),
```
vishwacsena: Updated spec. Make this `region`

packages/lu/src/commands/luis/up.ts
```diff
@@ -0,0 +1,141 @@
...
+    environmentname: flags.string({ description: '[Optional] environment name to include in LUIS app name', default: 'dev' }),
```
vishwacsena: Updated spec. Make this `suffix`

packages/lu/src/commands/luis/up.ts
feich-ms: fixed

packages/lu/src/commands/luis/up.ts
feich-ms: fixed

packages/lu/src/commands/luis/up.ts
feich-ms: fixed

feich-ms:
@vishwacsena, where is @microsoft/bf-luis-cli library published? How can I use it? Or do you mean importing it from local?

vishwacsena:
@feich-ms its [here](https://botbuilder.myget.org/feed/botframework-cli/package/npm/@microsoft/bf-luis)

@munozemilio can comment on when this will be pushed out to npm

feich-ms:
@vishwacsena I will switch to the new package once it is published to npm. Before that I will keep the old api. @munozemilio please let me know once it is published. Thanks.

packages/lu/src/parser/lubuild/core.ts
feich-ms: removed. No need.

packages/lu/src/parser/lubuild/core.ts
feich-ms: removed. no need.

vishwacsena:
@feich-ms now on npm - https://www.npmjs.com/package/@microsoft/bf-luis-cli

packages/lu/package.json
feich-ms: switch to @microsoft/bf-luis-cli

packages/lu/package.json
```diff
@@ -45,6 +45,7 @@
   },
   "dependencies": {
     "@microsoft/bf-cli-command": "1.0.0",
+    "@microsoft/bf-luis-cli": "4.7.0",
```
munozemilio: This should be 1.0.0

packages/lu/src/commands/luis/build.ts
```diff
@@ -0,0 +1,139 @@
...
+export default class LuisBuild extends Command {
+  static description = 'Build lu files to train and publish luis applications'
+
+  static examples = [`
+    $ bf luis:build --in {INPUT_FILE_OR_FOLDER} --authoringkey {AUTHORING_KEY} --botname {BOT_NAME} --dialog {true}
+  `]
+
+  static flags: any = {
+    help: flags.help({ char: 'h' }),
+    in: flags.string({ char: 'i', description: 'Lu file or folder', required: true }),
```
munozemilio: Most of the commands which require input support read the input from stdin...can we add this when passing a single file?

packages/lu/src/commands/luis/build.ts
```diff
@@ -0,0 +1,139 @@
...
+    flags.culture = flags.culture && flags.culture !== '' ? flags.culture : 'en-us';
+    flags.region = flags.region && flags.region !== '' ? flags.region : 'westus';
+    flags.suffix = flags.suffix && flags.suffix !== '' ? flags.suffix : 'development';
+    flags.fallbacklocale = flags.fallbacklocale && flags.fallbacklocale !== '' ? flags.fallbacklocale : 'en-us'
+ 
+    const luContents: Array<Content> = [];
+    let multiRecognizerDialogPath: string = '';
+    let luisSettingsPath: string = '';
+
+    process.stdout.write('Start to load lu files\n');
```
munozemilio: use this.log

packages/lu/src/commands/luis/build.ts
```diff
@@ -0,0 +1,139 @@
...
+    let files = await fileHelper.getLuFiles(flags.in, true, fileExtEnum.LUFile);
+
+    for (const file of files) {
+      let fileCulture: string;
+      let fileName: string;
+      const luFiles = await fileHelper.getLuObjects(undefined, file, true, fileExtEnum.LUFile);
+      const result = await LuisBuilder.build(luFiles, true, flags.culture);
+      const fileContent = result.parseToLuContent();
+      process.stdout.write(`${file} loaded\n`);
```
munozemilio: this.log

packages/lu/src/commands/luis/build.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    let multiRecognizerFileContent: any = {
+      "$type": "Microsoft.MultiLanguageRecognizer",
+      "recognizers": {
+      }
+    }
+
+    if (fs.existsSync(multiRecognizerDialogPath)) {
+      multiRecognizerFileContent = JSON.parse(await fileHelper.getContentFromFile(multiRecognizerDialogPath));
+      process.stdout.write(`${multiRecognizerDialogPath} loaded\n`);
```
munozemilio: this.log

packages/lu/src/parser/lubuild/LuisRecognizer.ts
```diff
@@ -0,0 +1,50 @@
+export class LuisRecognizer {
```
munozemilio: Should this be inside the Luis folder?

packages/lu/src/parser/lubuild/LuisSettings.ts
```diff
@@ -0,0 +1,21 @@
+export class LuisSettings  {
```
munozemilio: Luis folder?

packages/lu/src/parser/lubuild/core.ts
```diff
@@ -0,0 +1,297 @@
...
+
+        contents.push(luisSettingsContent);
+
+        let dialogFileContent = new DialogFileContent(contents);
+
+        return dialogFileContent;
+    }
+
+    private static async createApplication(currentApp: any, config: LUISConfig, client: any, recognizer: LuisRecognizer, name: string, culture: string): Promise<boolean> {
+        process.stdout.write(`Creating LUIS.ai application: ${name} version:0.1\n`);
```
munozemilio: We pushed so hard when moving Ludown to bf cli to keep responsibilities separated. CLI functionality (like console logging) should be at the CLI layer (command class). Parser should work as a consumable library to any application.

packages/lu/src/parser/lubuild/core.ts
```diff
@@ -0,0 +1,297 @@
+import {LUISConfig, Content, DialogFileContent} from './types'
```
munozemilio: If this builds a lu file, should this be in the lu folder?

packages/lu/package.json
feich-ms: fixed

packages/lu/src/commands/luis/build.ts
feich-ms: fixed

packages/lu/src/commands/luis/build.ts
feich-ms: fixed

packages/lu/src/commands/luis/build.ts
feich-ms: fixed

packages/lu/src/parser/lubuild/LuisRecognizer.ts
feich-ms: Sorry, the class name is ambiguous. Actually this class is not related with luis or lu, it is dialog asserts lubuild generated, so I will still keep it in this folder but changed it name to Recognizer instead of LuisRecognizer.

packages/lu/src/parser/lubuild/LuisSettings.ts
feich-ms: Sorry, the class name is ambiguous. Actually this class is not related with luis or lu, it is dialog asserts lubuild generated, so I will still keep it in this folder but changed it name to Settings instead of LuisSettings.

packages/lu/src/parser/lubuild/core.ts
feich-ms: Fixed. I split the logic in core.ts to several public functions and conbine these functions in build.ts. In this way, I can log all the messages before every function call.

packages/lu/src/parser/lubuild/core.ts
feich-ms: I removed LuisConfig and DialogFileContent class which are not very meaningful and then I exposed Content class in standalone file content.ts. Content class is not related to lu or luis concepts but just a common container which has id, path, content, name properties.

packages/lu/src/commands/luis/build.ts
feich-ms: fixed

feich-ms:
@munozemilio Comments resolved. Please let me know if you have further suggestions.

packages/lu/src/commands/luis/build.ts
```diff
@@ -0,0 +1,265 @@
...
+export default class LuisBuild extends Command {
+  static description = 'Build lu files to train and publish luis applications'
+
+  static examples = [`
+    $ bf luis:build --in {INPUT_FILE_OR_FOLDER} --authoringkey {AUTHORING_KEY} --botname {BOT_NAME} --dialog {true}
+  `]
+
+  static flags: any = {
+    help: flags.help({char: 'h'}),
+    in: flags.string({char: 'i', description: 'Lu file or folder', required: true}),
```
munozemilio: If the required property is left as true, even when passing data through stdin the command will ask for the --in to be passed. You can take a look at luis:convert where we have a similar behavior and we enforce the validation like this:
```
  if (!stdin && !flags.in) {
    throw new CLIError('Missing input. Please use stdin or pass a file location with --in flag')
  }
```

packages/lu/src/commands/luis/build.ts
```diff
@@ -0,0 +1,265 @@
+/*!
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import {Command, flags, CLIError} from '@microsoft/bf-cli-command'
+import {LuBuildCore} from './../../parser/lubuild/core'
+import {Content} from './../../parser/lubuild/content'
```
munozemilio: Whats the difference between Lu and Lubuild Content? Should Content extend LU?

packages/lu/src/commands/luis/build.ts
```diff
@@ -0,0 +1,265 @@
...
+    const apps = await luBuildCore.GetApplicationList(client)
+
+    while (luContents.length > 0) {
+      const subLuContents = luContents.splice(0, luisApiQps)
+      await Promise.all(
+        subLuContents.map(async content => {
+          let locale = content.Locale || flags.culture
+          let appName = ''
+
+          let currentApp = await luBuildCore.ParseLuContent(content.Content, locale)
```
munozemilio: If LuBuild extends or replaced by LU this can go away by using parseToLuis()?

packages/lu/src/commands/luis/build.ts
```diff
@@ -0,0 +1,265 @@
...
+
+    this.log('Start to handle applications\n')
+
+    const defaultLuisSchemeVersion = '4.0.0'
+    const luBuildCore = new LuBuildCore()
+    const client = luBuildCore.InitLuisClient(flags.authoringkey, `https://${flags.region}.api.cognitive.microsoft.com`)
+
+    const apps = await luBuildCore.GetApplicationList(client)
+
+    while (luContents.length > 0) {
```
munozemilio: This while condition is quite big and complex, if possible can we please break this into smaller functions with meaningful names so anyone can understand the purpose of it and debug if necessary?
kind of the refactor done in /src/parser/luis/luisValidator.js or /src/parser/luis/luConverter.js would be nice

packages/lu/src/commands/luis/build.ts
feich-ms: Good catch. I removed the required setting for --in and added above check in code.

packages/lu/src/commands/luis/build.ts
feich-ms: Nice suggestion. I looked at Lu definition and found that the lubuild Content definition just have two more properties of 'name' and 'path' than Lu definition, so I extended the lu class in my content definition and only added above two more properties. This really reduced the code repetition.

packages/lu/src/commands/luis/build.ts
feich-ms: yes. I used the parseToLuis() after I import lu definition in lu.js.

packages/lu/src/commands/luis/build.ts
feich-ms: Good suggestion. It is not friendly to read and debug the condition code piece. I break it into several small functions including 
InitApplicationFromLuContent()
UpdateApplication()
CreateApplication()
TrainAndPublishApplication()

feich-ms:
Hi @munozemilio thanks for your good suggestions. I have resolved them based on your comments. Pls let me know if they are not the right way.

packages/lu/src/parser/lubuild/core.ts
```diff
@@ -0,0 +1,212 @@
...
+        let matchedEntities = existingApp.entities.filter((x: any) => x.name === e.name)
+        if (matchedEntities && matchedEntities[0].children !== undefined) {
+          e.children = []
+        }
+      }
+    })
+
+    currentApp.name = existingApp.name
+
+    if (!this.isApplicationEqual(currentApp, existingApp)) {
```
munozemilio: return !this.isApplicationEqual(currentApp, existingApp)

packages/lu/package.json
```diff
@@ -45,13 +45,15 @@
   },
   "dependencies": {
     "@microsoft/bf-cli-command": "1.0.0",
+    "@microsoft/bf-luis-cli": "1.0.0",
```
munozemilio: If you are using this to get the azure sdk luis client I think it will be more efficient just to bring that dependency directly  "azure-cognitiveservices-luis-authoring": "2.1.1".

packages/lu/src/parser/lubuild/core.ts
feich-ms: good suggestion. fixed.

packages/lu/package.json
feich-ms: I looked at the 'azure-cognitiveservices-luis-authoring' package in npm and found it will be duplicated soon and '@azure/cognitiveservices-luis-authoring' willbe its replacement. So I used the '@azure/cognitiveservices-luis-authoring' together with '@azure/ms-rest-azure-js' instead. It works well.

feich-ms:
@munozemilio thanks for the good suggestions. Comments resolved.

PR #425, opened 2019-12-12
Nigao/testschema

ninggao:


packages/dialog/templates/numberSetdimension.dialog.lg
```diff
@@ -0,0 +1,20 @@
...
+{
+    "$schema": "@{appSchema}",
+    "$kind": "Microsoft.OnSetProperty",
+    "property": "@{property}",
+    "entity": "dimension",
+    "actions": [
+        {
+            "$kind": "Microsoft.SetProperty",
+            "property": "$@{property}",
+            "value": "@dimension"
```
chrimc62: This only works if property has the same properties as dimension.  Does it?

packages/dialog/templates/standard.schema
```diff
@@ -34,6 +35,25 @@
...
+                }
+            },
+            "required":["number", "units"],
+            "$templates": [
+                "money.lg",
+                "money.lu",
+                "objectSetmoney.dialog"
+            ]   
         }
     },
```
chrimc62: For these things, I think it makes more sense to put them in seperate schema files.  Otherwise every schema is going to include them whether or not they use money.

packages/dialog/templates/stringSet.dialog.lg
```diff
@@ -10,57 +10,9 @@
...
-                {
-                    "$kind": "Microsoft.SetProperty",
-                    "property": "$@{property}",
-                    "value": "turn.recognized.entities[turn.schema.properties['@{property}'].$mappings[0]]"
-                }
-            ]
+            "$kind": "Microsoft.SetProperty",
+            "property": "$@{property}",
+            "value": "turn.recognized.entities[turn.schema.properties['@{property}'].$mappings[0]]"
         }
```
chrimc62: This is not necessary in an OnSetProperty.  This will also not work in that $mappings is not defined for every property.  I suspect this whole thing should be dropped in favor of more specific ones.

chrimc62:
> - @{schemaName}-@{property}Setmoney.dialog

This whole file should go away once you map to a specific entity, right?

---
Refers to: packages/dialog/templates/stringSetmoney.dialog.lg:3 in 09c6c61. [](commit_id = 09c6c61d5f8aa998895ed0adb304def4792eea95, deletion_comment = False)

packages/dialog/templates/en-us/datetime.lu.lg
```diff
@@ -2,6 +2,6 @@
 - @{entityFilename()}
 
 # template
-- $PREBUILT:datetime @{roles()}
+- $PREBUILT:datetimeV2 @{roles()}
```
chrimc62: >datetimeV2 [](start = 12, length = 10)

We need to support both datetime (in some languages) and datetimeV2.

packages/dialog/test/commands/dialog/forms/address.schema
```diff
@@ -0,0 +1,27 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "definitions":{
```
chrimc62: >definitions [](start = 5, length = 11)

I think you can remove definitions and just use it as a property at the top.  We should talk.

packages/dialog/templates/numberSetdimension.dialog.lg
ninggao: Yes. This will be gone since dimension is also a composite entity (number, units). For dimension, we should do the same thing with money.

packages/dialog/templates/standard.schema
ninggao: This is some left over changes from my previous tries. I have removed the whole "money" definition.

packages/dialog/templates/stringSet.dialog.lg
ninggao: I agree. Now we don't really use it anymore. I will just drop it.

packages/dialog/templates/en-us/datetime.lu.lg
ninggao: Let's look back to this once new lubuild is ready.

packages/dialog/test/commands/dialog/forms/address.schema
ninggao: I already removed this.

PR #430, opened 2019-12-13
Nigao/moneyschema

ninggao:


packages/dialog/templates/money.schema
```diff
@@ -0,0 +1,17 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "money": {
+        "type": "money",
+        "$mappings": ["money"],
```
chrimc62: This is wrong.  It needs to be type: object.

packages/dialog/templates/moneyAsk.dialog.lg
```diff
@@ -0,0 +1,23 @@
...
+            "expectedProperties": [
+                "@{property}"
+            ]
+        },
+    ]
+}
+```
+
+# callAsk
+- @\{Ask@{property}()}
```
chrimc62: Since all of these look the same, seems like having a generic ask.dialog.lg to use in explicit templates would make sense rather than having a separate one for every type.  It does probably still make sense to do that for the auto-generated names though.

packages/dialog/templates/moneySet.dialog.lg
```diff
@@ -0,0 +1,25 @@
...
+{
+    "$schema": "@{appSchema}",
+    "$kind": "Microsoft.OnSetProperty",
+    "property": "@{property}",
+    "entity": "money",
+    "actions": [
+        {
+            "$kind": "Microsoft.SetProperty",
+            "property": "$@{property}.number",
+            "value": "@money['number']"
```
chrimc62: >            "value": "@money['number']"     [](start = 0, length = 43)

I'll check into why @money.number does not work--it should.

packages/dialog/test/commands/dialog/forms/sandwich.schema
```diff
@@ -17,7 +17,10 @@
             "$units": "inches"
         },
         "Name": {
-            "type": "string"
+            "type": "string",
+            "$mappings": [
+                "personName"
+            ]
```
chrimc62: Add utterance as well.

packages/dialog/templates/money.schema
ninggao: revised.

packages/dialog/test/commands/dialog/forms/sandwich.schema
ninggao: added.

PR #435, opened 2019-12-16
Nigao/moneyschema

ninggao:


packages/dialog/templates/dimensionSet.dialog.lg
```diff
@@ -0,0 +1,25 @@
...
+    "actions": [
+        {
+            "$kind": "Microsoft.SetProperty",
+            "property": "$@{property}.number",
+            "value": "@dimension['number']"          
+        },
+        {
+            "$kind": "Microsoft.SetProperty",
+            "property": "$@{property}.units",
+            "value": "@dimension['units']"
```
chrimc62: >['units'] [](start = 32, length = 9)

Still need to look into why .units doesn't work.

PR #454, opened 2019-12-25
change section id to specific pattern

feich-ms:
Composer team is working on managed lu feature. They need to do some CRUD operations to lu file content based on sections parsed by lu parser. Previously, we will regenerate the section id with GUID if section is updated, which makes it difficult to manage the section in composer side as id is changed. So this PR is to change section id to specific solid naming pattern. The id will be generated based on below rules:
1. intent/entity: type_name
2. import: type_path
3. modelInfo(comment): type_modelInfoText
4. qna(will be duplicated soon): type_questionListString

After these changes, the section id will keep unchanged after you update the section. But the id will be changed if you update the section with name changed.

We also expose the sectionHandler module in index.js to make all the needed resources availabe to composer


packages/lu/src/parser/lufile/qnaSection.js
```diff
@@ -15,7 +14,7 @@ class QnaSection {
         this.FilterPairs = this.ExtractFilterPairs(parseTree);
         this.Answer = this.ExtractAnswer(parseTree);
         this.Errors = [];
-        this.Id = uuidv4();
+        this.Id = `${this.SectionType}_${this.Questions.toString()}`;
```
Danieladu: we should double-check this one to avoid such result: `[object Object],[object Object]`

feich-ms: updated. changed it to this.Questions.join('_'). Here this.Questions is a string array

boydc2014:
Please also include the motivation of this change, the scenario requirement from Composer side, which could be helpful to reviewers.

boydc2014:
@munozemilio I approved this because from Composer side, we need a section id to be `stable`, so that we can continue update certain sections, please help confirm this meet your requirement or not. Thanks

PR #457, opened 2019-12-30
Bf cli luis functional tests

jspruance:


jspruance:
Functional tests take the subscription key from an env variable which can be provided like this:

export SUBSCRIPTION_KEY=99999 (linux)
set SUBSCRIPTION_KEY=99999 (windows)

then run 'rushx test'

if no sub key is provided, tests should log a message and exit

PR #464, opened 2020-01-09
add Luis TestDevSpec

hibrenda:
spec to bring bf lu test


scheyal:
Looks good. I'll add a --verbose flag or some other flag that permit showing less info, essentially just failed items and summary.
Also subscriptionKey term may change. LUIS terminology is under review. You can use this until we settle on final forms. 
I'll also add LUIS to this review.

specs/LUIS-TestDevSpec.md
```diff
@@ -0,0 +1,176 @@
...
+- LUIS model has been published to remote LUIS
+- LUIS subscriptionKey,  applicationId (trained model)
+- Testdata file
+
+### Summary
+
+- Support lu file format and json file format as the input file.
+- Support both intent and entity test.  
+- Support intent only test.
+- Entity types include the simple entity, composite entity, and the entity with role.
```
hazemelh: Have to add ML (with and without) structured entities as well.

specs/LUIS-TestDevSpec.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Prerequisites
+
+- LUIS model has been published to remote LUIS
+- LUIS subscriptionKey,  applicationId (trained model)
+- Testdata file
+
+### Summary
+
+- Support lu file format and json file format as the input file.
```
hazemelh: How do you expect to do this for lu files? If you are hitting the endpoint (which I assume because a prerequisite is that the model is published), then you need to extract the plain utterance from the lu formatted utterance.

specs/LUIS-TestDevSpec.md
```diff
@@ -0,0 +1,176 @@
...
+
+```
+
+USAGE
+  $bf luis:test
+
+OPTIONS
+  -a, --appid=appid                       A luis app id where to test the model
+  -h, --help                              luis:convert help
+  -i, --in=in                             Source .lu file(s) or LUIS application JSON model
```
hazemelh: not an application json in the input, probably a test application json, which is the same as the utterances section in the application json.

specs/LUIS-TestDevSpec.md
```diff
@@ -0,0 +1,176 @@
...
+- Utterance {@compositeType={@entityType1=foo} and {@entiyType2=bar}}
+
+Output: result.lu  
+
+> Total passed utterance: 8/10
+
+> <list out all the failed results>
+```
+
+Scenario 1: User’s expected intent does not match with the predict intent, list out the actual predicted top1 intent. (the predicted entity is same with the expected entity)
```
hazemelh: You will probably need to include the utterance itself somewhere so the user knows which utterance failed or passed instead of looking back at the original test file

specs/LUIS-TestDevSpec.md
```diff
@@ -0,0 +1,176 @@
...
+```
+Example:
+"Hello?"  (expected intent "Welcome")
+"I want to book the ticket." (expected intent "bookflight")
+"Check my Email." (expected intent "Email")
+```
+
+## Design Spec
+
+Based on the requirement, user needs to provide the testdata file and the LUIS configuration. This testdata includes the testing utterances labeled with user’s expected intent/entities. User will get a test report for the actual predicted results by running this command.
```
hazemelh: I'd also add Precision, Recall, and F1 for each intent and entity at the top so they can quickly get an at a glance performance results.

**IntentA**
Precision: 0.86
Recall: 0.54
F1: 0.77

specs/LUIS-TestDevSpec.md
hibrenda: Your understanding is correct. we leverage bf cli lu capability to parse lu format file into json format , then we can extract plain utterance from json format for test.

specs/LUIS-TestDevSpec.md
hibrenda: updated.

specs/LUIS-TestDevSpec.md
hibrenda: updated

specs/LUIS-TestDevSpec.md
hibrenda: updated

specs/LUIS-TestDevSpec.md
```diff
@@ -0,0 +1,179 @@
...
+- LUIS model has been published to remote LUIS
+- LUIS subscriptionKey,  applicationId (trained model)
+- Testdata file
+
+### Summary
+
+- Support lu file format and json file format as the input file.
+- Support both intent and entity test.  
+- Support intent only test.
+- Entity types include the simple entity, ml entity, composite entity, and the entity with role.
```
vishwacsena: LUIS today support the following entity types. Why can we not support them all? 
In luis.ai
- simple
- regex
- prebuilt 
- pattern.any
- list
- composite

In preview.luis.ai
- ml (flat or hierarchical)
- prebuilt
- regex
- pattern.any
- list
- composite (but this will be deprecated in the future)

specs/LUIS-TestDevSpec.md
```diff
@@ -0,0 +1,179 @@
...
+- LUIS subscriptionKey,  applicationId (trained model)
+- Testdata file
+
+### Summary
+
+- Support lu file format and json file format as the input file.
+- Support both intent and entity test.  
+- Support intent only test.
+- Entity types include the simple entity, ml entity, composite entity, and the entity with role.  
+- Support batch testing.
```
vishwacsena: Batch testing is not a public API. How do we plan to support this?

specs/LUIS-TestDevSpec.md
```diff
@@ -0,0 +1,179 @@
...
+- Result includes the summary at the beginning and the detail attached right after that.
```
vishwacsena: We should support the standard --out <fileName> [--force] flags for the command to write out the output file as .lu (or any file extension user chooses.,

vishwacsena:
@tsuwandy @hcyang it would be great if you guys can review as well. Thanks!

specs/LUIS-TestDevSpec.md
hibrenda: ml, simple, composite and role are supported in current spec as these features are Machine-learned entity. but regex, patter.any, list and prebuilt entity are exact text match and not labeled in current lu file. so these kind of entities will be ignored in the test result.

specs/LUIS-TestDevSpec.md
hibrenda: As the batch testing public API is not ready yet. we plan to use predict API and concurrent settings for batch testing. Later  we will migrate to the batch public API once it is ready.

specs/LUIS-TestDevSpec.md
hibrenda: We support  [--force] flags. (you could reference line 60 for details. )

specs/LUIS-TestDevSpec.md
vishwacsena: But all non ml types can be labeled if they are used with a role. We should support that. Can you update the spec and clarify?

specs/LUIS-TestDevSpec.md
vishwacsena: Ok. Since we are using the prediction API, we should parallelize this to the extent allowed by the subscription key limits. https://azure.microsoft.com/en-us/pricing/details/cognitive-services/language-understanding-intelligent-services/

specs/LUIS-TestDevSpec.md
hcyang: This "test" package primarily is to test ML model quality. Non-ML-type (rule-based) entity detection shall not fail, otherwise it's a LUIS system bug, not a quality issue.

specs/LUIS-TestDevSpec.md
hcyang: When we implemented the C#-based BotBuilder-Tools Dispatch, we had already tried the multi-threading "parallel" strategy in order trying to shorten the predicting time for a batch of queries. However, LUIS Service' throttle mechanism can fail many queries even with a small number of threads as LUIS Service's endpoint latency is usually 50ms per query, translated into 20 TPS, only about  half of the 50 TPS limit. I.e., a theoretic safe parallelism is 2, but latency fluctuation might still invoke throttling, cause endpoint-failure and retries. In short, unless things change, IMHO, it is not worth the effort.

hibrenda:
> This spec combines batch querying LUIS endpoint hits and creating an evaluation report meant for human consumption.
> These two should be split into two specs: a) Batch LUIS querying as well as generating a score file, and b) evaluating a generic, machine parseable score file and creating reports. I believe this spec should focus on the former, while the latter can already be handled by the "bf-dispatcher" library.
> For the a) spec, this package can evolve beyond LUIS to other NLP system like QnA Maker or Orchestrator.

So following report in summary could be covered by bf-dispatcher. is this correct ? @hcyang 
"> IntentA : Precision: 0.86, Recall: 0.54, F1: 0.77"
"> EntityA : Precision: 0.86, Recall: 0.54, F1: 0.77"

specs/LUIS-TestDevSpec.md
hibrenda: Thanks for sharing. we may not use "parallel" strategy but we still want to provide "batch testing" feature. we may using prediction API sequentially.

specs/LUIS-TestDevSpec.md
hibrenda: updated

hcyang:
> r

hcyang:
> > This spec combines batch querying LUIS endpoint hits and creating an evaluation report meant for human consumption.
> > These two should be split into two specs: a) Batch LUIS querying as well as generating a score file, and b) evaluating a generic, machine parseable score file and creating reports. I believe this spec should focus on the former, while the latter can already be handled by the "bf-dispatcher" library.
> > For the a) spec, this package can evolve beyond LUIS to other NLP system like QnA Maker or Orchestrator.
> 
> So following report in summary could be covered by bf-dispatcher. is this correct ? @hcyang
> "> IntentA : Precision: 0.86, Recall: 0.54, F1: 0.77"
> "> EntityA : Precision: 0.86, Recall: 0.54, F1: 0.77"

Yes, the bf-dispatcher package can create those metrics and much more.
One example of using the bf-dispatcher report library is in $/packages/dispatcher/src/model/evaluation/report/AppThresholdReporter.ts
Let me know if you have any questions.

hibrenda:
> > > This spec combines batch querying LUIS endpoint hits and creating an evaluation report meant for human consumption.
> > > These two should be split into two specs: a) Batch LUIS querying as well as generating a score file, and b) evaluating a generic, machine parseable score file and creating reports. I believe this spec should focus on the former, while the latter can already be handled by the "bf-dispatcher" library.
> > > For the a) spec, this package can evolve beyond LUIS to other NLP system like QnA Maker or Orchestrator.
> > 
> > 
> > So following report in summary could be covered by bf-dispatcher. is this correct ? @hcyang
> > "> IntentA : Precision: 0.86, Recall: 0.54, F1: 0.77"
> > "> EntityA : Precision: 0.86, Recall: 0.54, F1: 0.77"
> 
> Yes, the bf-dispatcher package can create those metrics and much more.
> One example of using the bf-dispatcher report library is in $/packages/dispatcher/src/model/evaluation/report/AppThresholdReporter.ts
> Let me know if you have any questions.

Thanks for the comments. Removed above from spec. Meanwhile besides the summary report, a JSON format detail log will also be generated.

PR #465, opened 2020-01-09
[lubuild] cli param typo optimization, bug fixes and test optimization 

feich-ms:
Pls notice that most of the changed files are test case files. Logic changes only happened in three files : **build.ts/recognizer.ts/build.test.ts**
1. fix the bug of force overwriting existing files when --out set. After the fix, it will not overwrite existing files unless --force set

2. optimize CLI parameter typo to align with other packages
    authoringkey -> authoringKey
    botname -> botName
    fallbacklocale -> fallbackLocale

3. fix a bug of loading existing Recognizer files
4. add tests to cover more corner cases

5. add uuid package in dev dependency


packages/lu/src/commands/luis/build.ts
```diff
@@ -187,19 +196,28 @@ export default class LuisBuild extends Command {
...
 
     // write dialog asserts
+    const contents = luBuildCore.GenerateDeclarativeAssets(recognizers, Array.from(multiRecognizers.values()), settings)
     if (flags.dialog) {
-      const contents = luBuildCore.GenerateDeclarativeAssets(recognizers, Array.from(multiRecognizers.values()), settings)
       for (const content of contents) {
         if (flags.out) {
-          this.log(`Writing to ${content.path}\n`)
-          await fs.writeFile(path.join(flags.out, path.basename(content.path)), content.content, 'utf-8')
+          const outFilePath = path.join(path.resolve(flags.out), path.basename(content.path))
```
chrimc62: Would be simpler to just compute outpath based on flags.out and do the same code for both.

PR #466, opened 2020-01-09
Optimize section operator and lu parser for managed lu requirements

feich-ms:
Optimize several aspects of section operator api and lu parser:

1. Section operator api:
    - expose Body field in SimpleIntentSection and NestedIntentSection class to benefit lu editing in composer
    - fix bug in SectionOperator to return orginal LUResource when updating a section not found. Before fix, it returned the SectionOperator class itself
    - optimize Newline with os EOF

2. Lu parser:
    - optimize LUFile g4 files to remove uncessary newline token
    - add errorIntentString rule to catch utterance without '-' before it and then throw a more friendly error message for this error like 'Invalid intent body line, did you miss '-' at line begin'
    - fix a bug when parsing utterances. Before the change, utterance with whitespace in pattern definition like 'show my {    listType} list' will throw errors when publishing. After the fix, it will trim the whitespace.
   - ignore case sensitivity for hard coded tokens in lexer


feich-ms:
@munozemilio this PR optimize several issues for section operator and lu parser. Please let me know if you have any suggestions.

PR #471, opened 2020-01-11
Fix bug that would prevent creating first model.

chrimc62:


packages/lu/src/commands/luis/build.ts
```diff
@@ -164,7 +164,7 @@ export default class LuisBuild extends Command {
 
         // compare models to update the model if a match found
         // otherwise create a new application
-        if (recognizer.getAppId() !== '') {
+        if (recognizer.getAppId() && recognizer.getAppId() !== '') {
```
munozemilio: if (recognizer.getAppId()) should evaluate the same

PR #477, opened 2020-01-17
[Lubuild] Expose lu build core for composer

feich-ms:
Move the core functionality in cli build.ts to a core builder.ts file and expose the build function in builder.ts so that both cli class and composer can use the core build functionality. The whole lu build process include three parts. 
1. load lu contents and other existing dialog contents
2. build luis applications(update or create applications and then train and publish them)
3. write/update dialog asset contents to corresponding files

To resolve issues #518, I added the lubuild.js to expose the class Builder which is using some node specific modules such as path. The Builder class will be used by composer server which is also node environment. Modules exported in index.js will be used by composer client and they will not touch node specific modules.

This PR also resolved the issue #548 which was asking to refine some lubuild requests


packages/lu/src/commands/luis/build.ts
```diff
@@ -40,267 +30,9 @@ export default class LuisBuild extends Command {
...
-      }
-    } while (!done)
-    this.log('done\n')
-
-    // publish applications
-    this.log(`${recognizer.getLuPath()} publishing version=${recognizer.versionId}\n`)
-    await delay(delayDuration)
-    await luBuildCore.PublishApplication(recognizer.getAppId(), recognizer.versionId)
-    this.log(`${recognizer.getLuPath()} publishing finished\n`)
+    const builder = new Builder(this)
```
munozemilio: This looks a bit odd, seems like the functionality was designed for the CLI and not for another application to consume.

packages/lu/src/parser/lubuild/builder.ts
```diff
@@ -0,0 +1,287 @@
...
+    let dialogFilePath = process.cwd()
+
+    // luis api TPS which means 5 concurrent transactions to luis api in 1 second
+    // can set to other value if switched to a higher TPS(transaction per second) key
+    let luisApiTps = 5
+
+    // set luis call delay duration to 1100 millisecond because 1000 can hit corner case of rate limit
+    let delayDuration = 1100
+
+    if (flags.stdin && flags.stdin !== '') {
```
munozemilio: Will composer read from stdin? this is an operation that looks CLI related only.

packages/lu/src/parser/lubuild/builder.ts
```diff
@@ -0,0 +1,287 @@
...
+      this.logger.log('Load lu content from stdin')
+      luContents.push(new Content(flags.stdin, 'stdin', true, flags.culture, path.join(process.cwd(), 'stdin')))
+      multiRecognizers.set('stdin', new MultiLanguageRecognizer(path.join(process.cwd(), 'stdin.lu.dialog'), {}))
+      settings = new Settings(path.join(process.cwd(), `luis.settings.${flags.suffix}.${flags.region}.json`), {})
+    } else {
+      this.logger.log('Start to load lu files\n')
```
munozemilio: Will composer consume this? Will they bypass the message and show it to the user? Are they going to parse it and then process accordingly? I think it is easier for an application to consume a library that returns an object, boolean or something easier to parse.

packages/lu/src/parser/lubuild/builder.ts
```diff
@@ -0,0 +1,287 @@
...
+      dialogFilePath = flags.in.endsWith(fileExtEnum.LUFile) ? path.dirname(path.resolve(flags.in)) : path.resolve(flags.in)
```
lei9444: If I use flags.in as a directory, FileHelper will find all lu files under this path. But if I want to choose some files to publish, and some lu files are not used, I have to call this API several times. Can we replace the property or add an array property to the flags?  Do you have any ideas about this? @feich-ms @munozemilio @boydc2014

packages/lu/src/commands/luis/build.ts
feich-ms: I have split the build function into three parts:
1. LoadContents() which is loading lu contents from lu files and existing dialog assets such as recognizers, multiRecognizers and settings from existing dialog files or create default ones if there are no existing dialog files

2. build() which accepts all the contents as inputs and then update/create, train and publish luis applications and update the dialog assets as well

3. writeDialogAssets() which accepts the dialog asset contents returned by build() and write dialog assets to corresponding files based on the cli parameters.

For composer, it can consume all the three functions or only the core build function if it has its own resources loading process.

packages/lu/src/parser/lubuild/builder.ts
feich-ms: I have moved this to LoadContents() function and make stdin as optional parameter. It composer consume this function, the stdin can be skipped.

packages/lu/src/parser/lubuild/builder.ts
feich-ms: Per the new structure, the logger is replaced by a delegate. For CLI, the delegate can be log message. For composer, it can be any delegate function if they want to parse the messages or do anything else. It can also be an empty function that does nothing. I have split the build process to three functions and the new core build function will not handle file systems so that composer or any other consumers can define their resource loading method and the call build function to update/create applications.

packages/lu/src/parser/lubuild/builder.ts
feich-ms: @lei9444 per the new design, you can write you own resource loading function or you can call loadResource() multiple times to load your specific files. After you load all the contents, you can call build() function to publish you applications.

feich-ms:
@munozemilio My changes to resolve issues #516 and #518 may not be reasonable. Could you give some suggestions

feich-ms:
@munozemilio I applied your latest changes of lu-cli and updated the build.ts and its tests in lu-cli based on the refactoring in this PR. The rush posttest should pass now.

packages/lu/src/commands/luis/build.ts
```diff
@@ -27,278 +25,106 @@ export default class LuisBuild extends Command {
...
     out: flags.string({description: 'Output location'}),
-    culture: flags.string({description: 'Culture code for the content. Infer from .lu if available. Defaults to en-us'}),
+    defaultCulture: flags.string({description: 'Culture code for the content. Infer from .lu if available. Defaults to en-us'}),
     region: flags.string({description: 'LUIS authoring region'}),
     suffix: flags.string({description: 'Environment name as a suffix identifier to include in LUIS app name'}),
     force: flags.boolean({char: 'f', description: 'Force write dialog and settings files', default: false}),
     dialog: flags.boolean({description: 'Write out .dialog files', default: false}),
-    fallbackLocale: flags.string({description: 'Locale to be used at the fallback if no locale specific recognizer is found. Only valid if --dialog is set'})
+    fallbackLocale: flags.string({description: 'Locale to be used at the fallback if no locale specific recognizer is found. Only valid if --dialog is set'}),
+    config: flags.string({description: 'Path to config'}),
```
munozemilio: why not use the CLI config file for this?

packages/lu/src/commands/luis/build.ts
```diff
@@ -27,278 +25,106 @@ export default class LuisBuild extends Command {
...
-    const stdin = await this.readStdin()
-    if (!stdin && !flags.in) {
-      throw new CLIError('Missing input. Please use stdin or pass a file or folder location with --in flag')
+    flags.stdin = await this.readStdin()
+
+    let files: string[] = []
+    if (flags.config) {
+      const configFilePath = path.resolve(flags.config)
+      if (fs.existsSync(configFilePath)) {
+        const configObj = JSON.parse(await fileHelper.getContentFromFile(configFilePath))
```
munozemilio: cli config file? you can get the path from this.config.configDir, the file name is config.json. e.g.
```
path.join(this.config.configDir, 'config.json')
```

packages/lu/src/commands/luis/build.ts
munozemilio: the description and the name are a bit misleading, can we name this LU config?

feich-ms: renamed to luConfig

PR #478, opened 2020-01-17
[LU Parser] Optimize error messages

feich-ms:
Optimize error messages when '-' missing before entity, qna question, qna filter line. After changes, it will throw messages like
 'Invalid list entity line, did you miss \'-\' at line begin'
'Invalid QnA question line, did you miss \'-\' at line begin'
'Invalid QnA filter line, did you miss \'-\' at line begin'


packages/lu/src/parser/lufile/LUFileParser.g4
```diff
@@ -89,7 +89,7 @@ newEntityDefinition
     ;
 
 newEntityListbody
-    : (normalItemString newline)+
+    : ((normalItemString newline) | errorItemString)+
```
boydc2014: can we just make DASH optional instead of adding a new pattern?

feich-ms:
@munozemilio, Hi I'm optimizing some parser grammar to make the error messages more friendly. Pls help to review and let me know if you have any suggestions. Thanks, Fei

PR #497, opened 2020-01-27
Upgrade azure auth and runtime clients

jspruance:
…ccordingly


jspruance:
unit & functional tests passed after this change

PR #502, opened 2020-01-29
Add optional save flag to luis:application:import

jspruance:


jspruance:
Customers would like to use 'luis:application:import' to import a new app based on a json file, but it is difficult to programmatically retrieve the newly created app Id afterwards.

Proposed change: add an optional 'save' flag so that the appId can be saved to config and be readily available for use for downstream operations. This is the exact same functionality currently available in 'luis:application:create'.

packages/luis/README.md
```diff
@@ -43,6 +45,534 @@ OPTIONS
   -h, --help  LUIS command help
 ```
```
jspruance: hmm...the url correction script didn't seem to run here...looking into it

jspruance: fixed

packages/luis/README.md
```diff
@@ -9,6 +9,8 @@
 <!-- toc -->
 * [Commands](#commands)
 <!-- tocstop -->
+* [Commands](#commands)
```
munozemilio: Fix duplicated tags

packages/luis/src/commands/luis/application/import.ts
```diff
@@ -47,9 +48,35 @@ export default class LuisApplicationImport extends Command {
...
       const response = await client.apps.importMethod(JSON.parse(appJSON), options)
       this.log(`App successfully imported with id ${response.body}.`)
+      if (save) {
+        const config = {
+          appId: response.body,
+          endpoint,
+          subscriptionKey
+        }
+        const saveConfigResponse = await this.saveImportedConfig(config, configDir)
+        if (saveConfigResponse) this.log('Config settings saved')
```
munozemilio: this is always true

jspruance: fixed

packages/luis/README.md
jspruance: fixed

packages/cli/README.md
```diff
@@ -15,6 +15,1585 @@ $ npm install -g @microsoft/botframework-cli
 
 # Commands
 <!-- commands -->
+* [`bf `](#bf-)
```
munozemilio: is this duplicated too?

packages/luis/src/commands/luis/application/import.ts
```diff
@@ -47,9 +48,35 @@ export default class LuisApplicationImport extends Command {
     try {
       const response = await client.apps.importMethod(JSON.parse(appJSON), options)
       this.log(`App successfully imported with id ${response.body}.`)
+      if (save) {
+        const config = {
+          appId: response.body,
+          endpoint,
+          subscriptionKey
+        }
+        const saveConfigResponse = await this.saveImportedConfig(config, configDir)
```
munozemilio: Is this const used?

packages/luis/README.md
```diff
@@ -49,6 +49,534 @@ _See code: [src/commands/luis/index.ts](https://github.com/microsoft/botframewor
 
 Creates a new LUIS application
 
+```
```
munozemilio: Ideally the change should be only in the luis:application:import. this looks a bit odd, is this correct?

cleemullins:
@munozemilio Please resolve this today, as @jspruance turns into a pumpkin today at 5pm.

PR #515, opened 2020-02-04
Qika/test

chrimc62:


packages/dialog/package.json
```diff
@@ -38,12 +38,13 @@
     "version": "npm run doc:readme && git add README.md"
   },
   "dependencies": {
```
chrimc62: openapi-types is missing as a dependency so rush fails.

packages/dialog/test/commands/dialog/forms/sandwich.schema
```diff
@@ -2,68 +2,28 @@
     "$schema": "http://json-schema.org/draft-07/schema",
     "type": "object",
     "properties": {
-        "Quantity": {
+        "petId": {
```
chrimc62: Why did you change this file at all?

packages/dialog/sandwich.schema
```diff
@@ -0,0 +1,72 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
+    "type": "object",
+    "properties": {
+        "Quantity": {
+            "type": "number",
```
chrimc62: This file does not belong here.

packages/dialog/sandwich_ver1/sandwich-BeginDialog.dialog
```diff
@@ -0,0 +1,12 @@
+
+{
+    "$schema": "https://raw.githubusercontent.com/microsoft/botbuilder-dotnet/master/schemas/sdk.schema",
+    "$kind": "Microsoft.OnBeginDialog",
```
chrimc62: I don't understand why the sandwich_ver1 stuff is here at all--that is generated and does not belong in the source tree.

packages/dialog/src/library/dialogGenerator.ts
```diff
@@ -353,7 +353,11 @@ export async function generate(
             properties: schema.schema.$public,
             entities: schema.entityTypes(),
             triggerIntent: schema.triggerIntent(),
-            appSchema: metaSchema
+            appSchema: metaSchema,
+            swaggerApi: schema.schema.swaggerApi,
+            swaggerMethod: schema.schema.swaggerMethod,
+            swaggerResponse: schema.schema.swaggerResponse,
+            swaggerBody: schema.schema.swaggerBody
         }
```
chrimc62: I don't think this belongs here at all.  What we really need is a way to pass into the generator extra properties as a json file.  This is in addition to the generation time templates file.

packages/dialog/src/library/swaggerDialogGenerator.ts
```diff
@@ -0,0 +1,116 @@
...
+        type: 'string'
+      }
+    case 'array':
+      // todo: handle arry
+      return undefined
+    case 'boolean':
+      return {
+        type: 'number',
+        minimum: 0,
+        maximum: 1
```
chrimc62: Why isn't this just boolean?

packages/dialog/templates/CompleteSetConfirmation.dialog.lg
```diff
@@ -16,7 +16,12 @@
...
                 {
-                    "$kind": "Microsoft.EndDialog"
+                  "$type": "Microsoft.HttpRequest",
+                  "method": "@{swaggerMethod}",
+                  "url": "@{swaggerApi}",
+                  "responseType": "Json",
+                  "resultProperty": "@{swaggerResponse}",
+                  "body": @{swaggerBody}
                 }
             ],
```
chrimc62: Rather than modifying the existing ones, this should really be in there own files.  Through your http.schema file you can control which templates are used.

packages/dialog/templates/en-us/common.lg
```diff
@@ -146,7 +146,7 @@ Please choose a value for @{name(property)} from \[@{join(foreach(turn.dialogEve
     "body": [
         {
             "type": "TextBlock",
-            "text": "sandwich",
+            "text": "FormData",
             "size": "large"
```
chrimc62: Why was this changed?

packages/dialog/test/commands/dialog/forms/testSwagger.json
```diff
@@ -0,0 +1,832 @@
...
+                      },
+                      "$ref": "#/definitions/Tag"
+                  }
+              },
+              "status": {
+                  "type": "string",
+                  "description": "pet status in the store",
+                  "enum": ["available", "pending", "sold"]
+              }
+          },
```
chrimc62: Seems like there is enough information here to be able to generate a default card for the result.  You could bind each element to an element of each name with the string responses.  We might not know that photoUrls should be pictures, but we could automatically generate with strings.  In the case of an enum, we could also generate the mapping from the enum to the description--by default the enum name.

chrimc62:
A few things:
1) We want to go from a swagger file to a .schema file which can be passed through the generator.
2) I would propose bf dialog:generate:swagger to produce the .schema file.
3) I think we mainly have the right mechanisms in the .schema file to point to the required templates.  It should not be necessary to override the existing templates--just add new ones pointed to by the generate schema file.
4) We do have a way to have a runtime .lg file, but in addition seems like it would be useful to be able to pass in JSON information to make available to the generator.  I'm going to think about this a bit more.
5) I don't understand why all the sandwich_ver1 stuff or why the sandwich stuff was included or modified.

packages/dialog/src/commands/dialog/swagger.ts
```diff
@@ -0,0 +1,72 @@
...
+        outDir = ppath.join(schemaName + '-resources')
+      }
+      await gen.generate(schemaName, outDir, flags.schema, flags.locale, flags.templates, flags.force,
+        (type, msg) => {
+          if (type === gen.FeedbackType.message
+            || type === gen.FeedbackType.error
+            || (type === gen.FeedbackType.info && flags.verbose)) {
+            this.progress(msg)
+          }
+        })
```
chrimc62: We should put this into seperate steps.  That way you don't need to duplicate all of the flags and someone can modify the generated schema by hand before generating the bot.

packages/dialog/sandwich_ver1/sandwich-BeginDialog.dialog
coldplaying42: We tried to test and understand FormDialog based on the sandwich.schema before. All the stuff on sandwich dialog should not be uploaded. We will remove them.

packages/dialog/src/library/swaggerDialogGenerator.ts
coldplaying42: This is the hacking code to define the range.

packages/dialog/templates/CompleteSetConfirmation.dialog.lg
coldplaying42: We will add additional dialog.lg to generate HttpRequest dialog file.

packages/dialog/src/library/swaggerDialogGenerator.ts
chrimc62: Yes, but you should use the boolean type rather than a number, right?

---
In reply to: [374461561](https://github.com/microsoft/botframework-cli/pull/515#discussion_r374461561) [](ancestors = 374461561)

packages/dialog/templates/CompleteSetConfirmation.dialog.lg
```diff
@@ -9,7 +9,7 @@
         {
             "$kind": "Microsoft.SetProperty",
             "property": "$CompleteConfirmation",
-            "value": "@Confirmation"
+            "value": "=@Confirmation"
```
ninggao: This is fixed in the master. You can drop the change in this file.

packages/dialog/templates/CompleteSetConfirmationAndMakeHttpRequest.dialog.lg
```diff
@@ -0,0 +1,52 @@
...
+    "actions": [
+        {
+            "$kind": "Microsoft.SetProperty",
+            "property": "$CompleteConfirmation",
+            "value": "@Confirmation"
+        },
+        {
+            "$kind": "Microsoft.IfCondition",
+            "condition":"$CompleteConfirmation == 'no'",
+            "actions": [
```
ninggao: I think this should be in a separate Swagger folder, since it will override the current completeconfirmation-assign-confirmation.dialog. Also following the new name pattern, it should just be completeconfirmation-assign-confirmation.dialog.

packages/dialog/templates/HttpRequest.dialog.lg
```diff
@@ -0,0 +1,25 @@
+# template
```
ninggao: Since there are no properties involved, it should be library-HttpRequest.dialog.lg.

packages/dialog/templates/en-us/common.lg
```diff
@@ -193,6 +193,35 @@ Please choose a value for @{name(property)} from \[@{join(foreach(turn.dialogEve
 # PropertyToChange(value)
 - @{PROPERTYName(value)}
 
+# cardResponse(content)
+[Activity
```
ninggao: The new functions could be moved to a new file (library-Swagger for example). And then in swagger branch, it should be included in the standard.schema.

packages/dialog/templates/http.schema
```diff
@@ -0,0 +1,73 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema",
```
ninggao: Why this file looks so similar to our standard.schema?

packages/dialog/templates/stringSetstring.dialog.lg
```diff
@@ -9,28 +9,6 @@
     "property": "@{property}",
     "entity": "{property}Entity",
     "actions": [
```
ninggao: If you want to remove this, you should create a file in a separate Swagger folder so it doesn't override our standard schemas.

cleemullins:
Closing due to inactivity. @chrimc62 please re-open if this is a PR we still need to take.

PR #520, opened 2020-02-04
add number validation

ninggao:


packages/dialog/templates/numberSetnumber.dialog.lg
```diff
@@ -11,17 +11,33 @@
...
         {
-            "$kind": "Microsoft.SendActivity",
-            "activity": "@{callSetMessage()}"
-        },
-        {
-            "$kind": "Microsoft.SetProperty",
-            "property": "$@{property}",
-            "value": "=@number"          
+            "$kind": "Microsoft.IfCondition",
+            "condition":"or(and(dialogClass.schema.properties['@{property}'].minimum, less(@number, dialogClass.schema.properties['@{property}'].minimum)), and(dialogClass.schema.properties['@{property}'].maximum, greater(@number, dialogClass.schema.properties['@{property}'].maximum)))",
```
chrimc62: >dialogClass.schema.properties['@{property}'].minimum [](start = 32, length = 52)

I don't think you need this.  Less with null should be false.

packages/dialog/templates/numberSetnumber.dialog.lg
```diff
@@ -11,17 +11,33 @@
...
+            "actions":[
+                {
+                    "$kind": "Microsoft.SendActivity",
+                    "activity": "@{callNumberAlert()}"
```
chrimc62: >callNumberAlert [](start = 35, length = 15)

Hmmm.  Seems like we might be missing something here--a way to say do the same prompt again.  In what we have generated so far, prompts tend to be predicatable, i.e. we keep doing the same one until fulfilled.  If though someone used a Random secondary selector the prompt would be different.  We don't need to fix this here, but it is something to think about.

packages/dialog/templates/en-us/common.lg
```diff
@@ -24,9 +24,19 @@ Which value do you want for @{name(property)}?
...
 - Enter a number for @{name(property)} between @{dialogClass.schema.properties[property].minimum} and @{dialogClass.schema.properties[property].maximum}
+- ELSEIF:  @{dialogClass.schema.properties[property].minimum}
+- Enter a number for @{name(property)} that is greater than @{dialogClass.schema.properties[property].minimum}
+- ELSEIF: @{dialogClass.schema.properties[property].maximum}
+- Enter a number for @{name(property)} that is less than @{dialogClass.schema.properties[property].maximum}
 - ELSE:
 - Enter a number for @{name(property)}
 
+# numberValidation(property, number)
+- IF: @{dialogClass.schema.properties[property].minimum && less(number, dialogClass.schema.properties[property].minimum)}
```
chrimc62: >dialogClass.schema.properties[property].minimum  [](start = 8, length = 48)

Do you need this?

packages/dialog/templates/numberSetnumber.dialog.lg
ninggao: Removed.

packages/dialog/templates/numberSetnumber.dialog.lg
ninggao: Agree. After check/alert/validation/help, we should have a way to let the users retry.

packages/dialog/templates/en-us/common.lg
ninggao: Tested, and removed.

PR #521, opened 2020-02-04
fix file content unit tests, add entries to index.ts

hcyang:
Fix a file size comparison unit test that fails on Mac OS/Linux machines.
Add Report entries into index.ts.


vishwacsena:
Verified this works on MacOS.

PR #536, opened 2020-02-05
ludown content is not supported in import command

in4margaret:
` bf luis:application:import --endpoint https://{region}.api.cognitive.microsoft.com/ --subscriptionKey yourKey --name AppName --in ./ludown.lu ` from the code it looks like it wasn't intended to work with ludown files


munozemilio:
The documentation is autogenerated from each plugin, this needs to be addressed in the Luis plugin and then propagated to the CLI doc. Also the LU file should be supported

PR #551, opened 2020-02-07
Silent ci install

### LINKED ISSUE #504: Update checker test should avoid touching git track files
From Hung-Chih:

> The   "lastversioncheck” timestamp entry in the file “packages/cli/test/fixtures/lastversioncheck/package.json” can change from time to time after a some Rush command.

This is indeed a weird thing where you have to be careful not to check this in after you run a test. We should fix it so it doesn't touch the file in the test.

### LINKED ISSUE #517: Microsoft telemetry Y/N prompt breaking azure pipeline
I have been setting up azure pipeline where after installing the various dependencies on the build agent (botframework-cli etc) and running command `bf luis:convert` in the script appears to display the following prompt -

`Help us improve products by allowing Microsoft to collect anonymous command and flags usage: (Y/N):` 

Is there a workaround for this to supress on azure hosted agents?

munozemilio:
Fixes #517 
Fixes #504

### DIFF
--- a/packages/cli/scripts/postinstall.js
+++ b/packages/cli/scripts/postinstall.js
@@ -3,6 +3,7 @@ const {cli} = require('cli-ux')
 const chalk = require('chalk')
 const path = require('path')
 const fs = require('fs-extra')
+const isCI = require('is-ci')
 const pjson = require('../package.json');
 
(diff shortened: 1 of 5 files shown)


packages/cli/scripts/postinstall.js
```diff
@@ -64,7 +68,9 @@ const getUserConfig = async () => {
 }
 
 const init = async () => {
-  await promptTelemetry()
+  if (!isCI) {
```
scheyal: Reminder: Remove?

PR #554, opened 2020-02-08
bf-lu perf updates

vishwacsena:
Made bunch of fixes to improve perf for core bf-lu operations. 

Notable example include large .lu files - e.g. email skill.lu
This multi-file convert now takes ~1,715ms .vs. 5,704ms, better by about 70%
> time bin/run luis:convert --in emailskills.lu
real   0m1.715s           [old: 0m5.704s]
user    0m2.056s .     [old: 0m6.057s]
sys     0m0.179s         [old: 0m0.182s]

The BF-LU tests now take ~89s .vs. ~129s, better by about 30%. 
> time npm run test 
real    1m28.603s     [old: 2m8.290s]
user    0m48.068s   [old: 1m38.726s]
sys     0m2.266s       [old: 0m3.117s]

### DIFF
--- a/packages/lu/src/parser/lufile/parseFileContents.js
+++ b/packages/lu/src/parser/lufile/parseFileContents.js
@@ -725,6 +725,7 @@ const parseAndHandleNestedIntentSection = function (luResource, enableMergeInten
 const parseAndHandleSimpleIntentSection = function (parsedContent, luResource) {
     // handle intent
     let intents = luResource.Sections.filter(s => s.SectionType === SectionType.SIMPLEINTENTSECTION);
+    if (!parsedContent.LUISJsonStructure.utteranceHash) parsedContent.LUISJsonStructure.utteranceHash = {}
     if (intents && intents.length > 0) {
         for (const intent of intents) {
             let intentName = intent.Name;
@@ -733,6 +734,7 @@ const parseAndHandleSimpleIntentSection = function (parsedContent, luResource) {
             for (const utteranceAndEntities of intent.UtteranceAndEntitiesMap) {
                 // add utterance
                 let utterance = utteranceAndEntities.utterance.trim();
+                let uttHash = helpers.hashCode(utterance);
                 // Fix for BF-CLI #122. 
                 // Ensure only links are detected and passed on to be parsed.
                 if (helpers.isUtteranceLinkRef(utterance || '')) {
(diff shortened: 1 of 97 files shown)


packages/lu/src/parser/lufile/parseFileContents.js
```diff
@@ -1538,15 +1551,15 @@ const handleClosedList = function (parsedContent, entityName, listLines, entityR
...
                     throw (new exception(retCode.errorCode.SYNONYMS_NOT_A_LIST, error.toString(), [error]));
                 }
-                if (!nvExists.list.includes(item)) nvExists.list.push(item);
+                nvExists.list.push(item);
             })
         }
     });
-
+    // de-dupe the list
+    if (nvExists && nvExists.list)nvExists.list = [...new Set(nvExists.list)];
```
feich-ms: whitespace before nvExists.list = [...new Set(nvExists.list)]

packages/lu/src/parser/luis/luis.js
```diff
@@ -62,25 +63,29 @@ class Luis {
...
         }
-        checkAndUpdateVersion(this)
+        helpers.checkAndUpdateVersion(this)
     }
 }
 
 module.exports = Luis
 
 const initialize = function(instance, LuisJSON) {
+    instance.utteranceHash = {};
```
munozemilio: Why adding this to the Luis object?

packages/lu/src/parser/luis/luis.js
```diff
@@ -104,6 +109,36 @@ const compareString = function(a, b) {
     return 0;
 }
 
+const mergeResultsWithHash = function (blob, finalCollection, type, hashProperty) {
+    if (blob[type].length === 0) { 
+        return
+    }
+    if (!finalCollection.utteranceHash) finalCollection.utteranceHash = {};
```
munozemilio: Why not making this local?

packages/lu/src/parser/lufile/parseFileContents.js
vishwacsena: Fixed

packages/lu/src/parser/luis/luis.js
vishwacsena: Updated this to be local.

vishwacsena:
@munozemilio updated. I also noticed that separating the qna from qna alterations causes all files to parsed twice which is expensive. We should probably have qnabuilder do both qnaList as well as qnaAlterations with a single parse.

packages/lu/src/parser/luis/luis.js
```diff
@@ -58,35 +59,38 @@ class Luis {
...
     }
 }
 
 module.exports = Luis
 
-const initialize = function(instance, LuisJSON) {
+const initialize = function(instance, LuisJSON, hashTable = undefined) {
     for (let prop in LuisJSON) {
         instance[prop] = LuisJSON[prop];
+        if (hashTable !== undefined && (prop === LUISObjNameEnum.UTTERANCE || prop === LUISObjNameEnum.PATTERNS)) {
```
munozemilio: This operation should be done for both scenarios, if Luis object is empty or already initialized.

vishwacsena: Fixed

PR #559, opened 2020-02-11
add numberSetdimention.dialog; remove unused enum from standard.schema

ninggao:


packages/dialog/templates/standard.schema
```diff
@@ -48,7 +48,6 @@
     "definitions": {
         "PROPERTYName": {
             "type": "string",
-            "enum": "@{properties}",
             "$entities": ["PROPERTYName"]
```
chrimc62: You cannot remove this.  Without this none of the property names will work--this is where they get copied from.  This will be the list of properties if you are synced to the latest. cli

packages/dialog/templates/en-us/common.lg
```diff
@@ -37,6 +37,12 @@ Which value do you want for @{name(property)}?
 - ELSEIF: @{greater(number, dialogClass.schema.properties[property].maximum)}
 - @{number} is greater than the maximum value @{dialogClass.schema.properties[property].maximum}.
 
+# unitsAlert(property)
+- IF: @{dialogClass.schema.properties[property].units}
```
chrimc62: >units [](start = 48, length = 5)

This needs to be $units.  JSON Schema supports adding only new properties that start with $.

Issue #119, opened 2019-09-10
Need MSLG PM Spec


scheyal:
We need a PM spec for MSLG. 

See Mini Spec template and Contribution guidelines in specs folder

vishwacsena:
Pushing out to R8.

PR #572, opened 2020-02-13
LG spec

vishwacsena:
Addresses #119


specs/LG_spec.md
```diff
@@ -0,0 +1,68 @@
...
+
+3. Expand
+```bash
+> bf lg:expand
+```
+Expand one or all templates in a .lg file or an inline expression.
+
+| Parameter                   | Description                                                                                                         |
+|-----------------------------|---------------------------------------------------------------------------------------------------------------------|
+| -i, --in <file or folder>   | [Required] .lg file or folder that contains .lg file.                                                               |
```
Danieladu: expand means expand the specific template in the specific lg file, If mslg accept lg file list,  means expand all the templates of all the lg files?  and the --templateName would lose effectiveness
-- | --

vishwacsena: Not necessarily. If --templateName is specified, then we should only expand that template by name across one or more .lg files pointed to via --in (either file or folder). 
If --all is specified, then we should expand all templates across one or more .lg files pointed to via --in.

vishwacsena:
@scheyal can I get your review as well for consistency of commands and args from BF CLI POV?

Issue #548, opened 2020-02-07
luis:build refinements


vishwacsena:
There are few improvements/ refinements to make to luis:build

1. Today, by default the write all generated files to the cwd() or the --out folder (if specified). We should change this behavior such that, **in cases where no --out folder is specified**,
    - any .dialog file generated should go to the respective folder where the .lu file was found. As an example, if the user ran `bf luis:build --in /myBot` and we found these two .lu files - `./dialog1/dialog1.lu` and `./dialog2/dialog2.lu` then all *.lu.dialog for dialog1.lu should go under `./dialog` and those for dialog2.lu should go under `./dialog2`.
   - **NOTE:** this is the default behavior when no --out folder is specified
2. luis:build write out a lot to console. we should move all these messages to be rendered only when `--log` is specified. By default, 
    - in cases where --dialog is not specified, write out the appIds list at the end of the operation
    - in cases where --dialog is specified, write `Successfully wrote .dialog files to <out-folder>`
    - write to console.error when there are any errors that prevent successful completion of the command
3. Replace `--culture` with `--defaultculture`
4. Add a `--config <path-to-config>`. This is the exact same [luconfig.json][1] that lubuild supported.
```jsonc
{
// CLI argument for --botName can be skipped if this is specified
    "name":"MyProject",
// CLI argument for --defaultCulture (--culture) can be skipped if this is specified. This is the application culture if a single .lu file is passed via --in. This is also the default culture for any .lu file without a lang-code in file name if a folder is passed in to --in. (e,.g. foo.lu will be foo.fr-fr.lu assuming --defaultCulture is set to fr-fr. This will continue to default to en-us
    "defaultLanguage":"en-us",
   // defaults to false when not specified. When set to true, old model version (when we create a new version) is automatically deleted for all applications
   "deleteOldVersion": true, 
   // absolute or relative path to set the output folder.
   "out": ".",
    // defaults to false when absent
    "writeDialogFiles": true,
    // each model is an application
    "models": [
        // each line is to an lu file and each line corresponds to a luis application. 
        "test/test.lu",
        "test/sample/x/Contoso.Foo.lu",
        "test/test.fr-fr.lu"
    ]
}
```

 
[1]:https://github.com/microsoft/botbuilder-tools/tree/V.Future/packages/lubuild#config-file

feich-ms:
try to fix in PR #477

vishwacsena:
@feich-ms ok. Let me know when you have it ready and I can test it.

[...]

PR #576, opened 2020-02-17
[Lubuild] fix multiple bugs: case sensitive issue, default version issue and exception not throw out issue

feich-ms:
Fix several bugs:
1. As known to us, values of utterances/patterns are automatically converted to lower case in luis portal even though typed words are upper case. When comparing app json from lu file with the one from remote existing luis app, we should to ignore cases. To achieve this, I converted the objects to string, then to lower case so that the comparison works insensitively.
2. Some properties in our luis definition don't align with the object pulled from luis portal. Fixed in this PR
3. Default version 0.1 was included by previous PR when no version is specified which caused some bugs when updating version. In this PR, version from lu will count only if it's value is over than existing app. For example, if the app with the same app id in luis portal is version 0.1, and the version from the lu file is 0.3, 0.3 > 0.1, then 0.3 will count, otherwise, 0.1 will auto increase by 0.1 to 0.2
4. Add test cases for above bug fixes
5. Add test cases for #548 
6. Fix bug #582


feich-ms:
@vishwacsena, sure, will add tests soon.

feich-ms:
Tests added @vishwacsena

packages/lu/test/commands/luis/build.test.ts
```diff
@@ -555,4 +555,36 @@ describe('luis:build update dialog assets successfully when dialog assets exist'
       expect(await compareFiles('./../../../results/foo.fr-fr.lu.dialog', './../../fixtures/testcases/lubuild/foo2/dialogs/foo.fr-fr.lu.dialog')).to.be.true
       expect(await compareFiles('./../../../results/foo.zh-cn.lu.dialog', './../../fixtures/testcases/lubuild/foo2/dialogs/foo.zh-cn.lu.dialog')).to.be.true
```
munozemilio: Please add this test to the Luis package since the command is now there

feich-ms: @munozemilio, fixed. moved it to luis package.

feich-ms:
@munozemilio, comments resolved. Also added test cases for luConfig support of #548. Fixed bug of #582 as well.

packages/lu/src/parser/lubuild/builder.ts
```diff
@@ -36,8 +37,18 @@ export class Builder {
...
       const luFiles = await fileHelper.getLuObjects(undefined, file, true, fileExtEnum.LUFile)
-      const result = await LuisBuilder.build(luFiles, true, culture)
-      const fileContent = result.parseToLuContent()
+
+      let fileContent = ''
+      try {
+        const result = await LuisBuilder.build(luFiles, true, culture)
+        fileContent = result.parseToLuContent()
+      } catch (err) {
+        if (err instanceof exception) {
```
munozemilio: Ideally all handled exceptions inside lu package should throw the utils/exception so the consumer can handle those

packages/luis/src/commands/luis/build.ts
```diff
@@ -38,93 +38,97 @@ export default class LuisBuild extends Command {
...
+        }
       } else {
-        this.log(`No changes to the .dialog files in ${outputFolder}`)
+        this.log('The published application ids:')
+        this.log(JSON.parse(dialogContents[dialogContents.length - 1].content).luis)
       }
-    } else {
-      this.log('The published application ids:')
-      this.log(JSON.parse(dialogContents[dialogContents.length - 1].content).luis)
+    } catch (error) {
```
munozemilio: Maybe not all errors should be thrown as CLIError, makes more sense to handle the exception thrown from the lu library as CLIError and throw everything else as unexpected error. We are doing the same in luis:convert.

packages/lu/src/parser/lubuild/builder.ts
feich-ms: Yes, fixed. Throw the error of exception type but added more details in the error.text to help locate the specific lu file in which errors occurred.

packages/luis/src/commands/luis/build.ts
feich-ms: Make sense, fixed. All errors of exception type are thrown as CLIError and others are thrown as unexpected errors.

feich-ms:
Comments resolved. Thanks @munozemilio

feich-ms:
@munozemilio could you give approval if no other issues for this PR. Thanks.

vishwacsena:
@feich-ms I tried this E2E and pushed up two sets of deltas up to your branch. I'm not familiar with nock and some of the lubuild tests are failing. Can you take a look and address?

vishwacsena:
@munozemilio except for addressing tests, it would be good if you get a review as well. Thanks!

packages/lu/src/parser/lubuild/builder.ts
```diff
@@ -36,16 +38,25 @@ export class Builder {
...
-      const fileContent = result.parseToLuContent()
+
+      let fileContent = ''
+      let result
+      try {
+        result = await LuisBuilder.build(luFiles, true, culture)
+        fileContent = result.parseToLuContent()
+      } catch (err) {
+        err.text = `Invalid LU file ${file}: ${err.text}`
+        throw err
```
munozemilio: throw exception

packages/lu/src/parser/lubuild/core.ts
```diff
@@ -9,8 +9,11 @@ import {Settings} from './settings'
 import {isEqual, differenceWith} from 'lodash'
 import {CognitiveServicesCredentials} from '@azure/ms-rest-azure-js'
 import {LUISAuthoringClient} from '@azure/cognitiveservices-luis-authoring'
+//const LUIS = require('./../luis/luis')
```
munozemilio: Delete commented imports

packages/lu/src/parser/luis/luis.js
```diff
@@ -35,11 +35,13 @@ class Luis {
...
-        let props = ['intents', 'composites', 'entities', 'closedLists', 'regex_entities', 'model_features', 'patternAnyEntities', 'prebuiltEntities']
-        for (const key of props) {
-            this[key].sort(sortComparers.compareNameFn)
+        for (const prop in this) {
+            if (prop !== 'utterances' && prop !== 'patterns' && Array.isArray(this[prop])) this[prop].sort(sortComparers.compareNameFn)
         }
         this.utterances.sort(sortComparers.compareIntentFn);
+        this.utterances.sort(sortComparers.compareUttTextFn);
+        this.patterns.sort(sortComparers.compareIntentFn);
+        this.patterns.sort(sortComparers.comparePatternFn)
```
munozemilio: is this intended?

packages/lu/src/parser/lubuild/builder.ts
vishwacsena: Fixed.

packages/lu/src/parser/lubuild/core.ts
vishwacsena: Fixed

packages/lu/src/parser/luis/luis.js
vishwacsena: Reverted.

feich-ms:
@munozemilio, tests failures are fixed. Please help to approve. Thanks.

feich-ms:
I added more exposed modules in index.js and composerindex.js in the latest commit. My intention is to expose modules that composer client/broswer consumes in composerindex.js and other modules in index.js. Parser and SectionHandler is consumed by composer client and luBuild is consumed by composer server. Please let me know if this is not the reasonable way. Thanks.

packages/lu/src/parser/index.js
```diff
@@ -27,10 +27,8 @@ const modules = {
 
         Parser: require('./lufile/classes/parserObject')
     },
-    sectionHandler: {
-        luParser: require('./lufile/luParser'),
-        sectionOperator: require('./lufile/sectionOperator'),
-        luSectionTypes: require('./utils/enums/lusectiontypes')
+    luBuild: {
```
munozemilio: Hi @feich-ms, the composerindex looks good but in this one I think our intention is to reduce our scope since the only consumer of this is composer, not the average dev. Composer can consume the lubuild index directly as we are doing in the commands

feich-ms: fixed. removed luBuild related changes

PR #583, opened 2020-02-18
Nigao/splitlibrary

ninggao:


packages/dialog/templates/en-us/enumEntity.lg.lg
```diff
@@ -3,7 +3,7 @@
 
 >>> This is because of issues around ' and } in strings--see enumEntityDesired.lg for a test case
 # template
-- @{propertyEntity()}\n@{chooseEntity()}\n\[library.lg]
+- @{propertyEntity()}\n@{chooseEntity()}\n\[library.lg]\n\[library-Choose.lg]
```
chrimc62: FYI, I have started replacing these multi-line things using ``` which makes clearer what is going on.

PR #595, opened 2020-02-21
Nigao/boolean

ninggao:


packages/dialog/templates/boolean-assign-boolean.dialog.lg
```diff
@@ -0,0 +1,29 @@
...
+            "property": "$@{property}",
+            "value": "=@boolean"          
+        }
+    ]               
+}
+```
+
+# callSetMessage
+- @\{setPropertyMessage('@{property}', @boolean)}
+
```
chrimc62: I suspect we could inline this now if we wanted to.

packages/dialog/templates/en-us/booleanEntity.lu.lg
```diff
@@ -0,0 +1,7 @@
+# filename
+- @{entityLUFilename()}
+
+# template
+- $@{property}Entity:boolean
```
chrimc62: >$@{property}Entity:boolean [](start = 2, length = 26)

I'm a little fuzzy on why we need this and it appears to be in the old syntax.

packages/dialog/templates/en-us/booleanProperty.lg.lg
```diff
@@ -0,0 +1,16 @@
+# filename
+- @{propertyFilename()}
+
+# template
+- @{AskTemplate()}\n@{NameTemplate()}\n@{ValueTemplate()}\n\[@{schemaName}-@{property}Entity.@{locale}.lg]\n\[library.lg]\n\[library-Missing.lg]
+
```
chrimc62: Multi-line with ```

packages/dialog/templates/en-us/library-Missing.lg
```diff
@@ -20,5 +20,11 @@ Which value do you want for @{name(property)}?
 Enter a value for @{name(property)}
 ```
 
+# askBoolean(property)
+- ```
+@{askHelp()}
+Enter a value for @{name(property)}: true or false
+```
```
chrimc62: Is this a little weird for all done?

ninggao: Ask is only triggered when a property's type is Boolean, and it's included in requiredProperties. CompleteConfirmation shouldn't be added to the requiredProperties list, so this Ask should never be triggered.

packages/dialog/templates/boolean-assign-boolean.dialog.lg
ninggao: I will make a separate PR since this change may involve all .dialog files.

packages/dialog/templates/en-us/booleanProperty.lg.lg
ninggao: Done

packages/dialog/templates/en-us/booleanEntity.lu.lg
ninggao: changed to new syntac.

PR #601, opened 2020-02-25
cross-train core library in bf-lu

feich-ms:
1. Rename luCrossTrainer.js to crossTrainer.js and expose cross-train.js to handle files input and output. Moved these two files to cross-train folder
2. Pull fathers's qna questions to children's interruption intents
3. config file of mapping rules should be specified by flags.config parameter. If not, it will read default config.json from the folder specified by --in
4. Remove duplicated utterances in interruption intent to make sure it has no duplications with other intent utterances in the lu file
5. Remove all duplications of lu utterances in a intent and qna questions in a qna section


feich-ms:
@vishwacsena working on the test refactoring. Will give updates when ready.

feich-ms:
@vishwacsena, cross train core library related tests are refactored. The PR looks much more clean now.

PR #605, opened 2020-02-25
Chrimc/dollar

chrimc62:
Update to use adaptive-expressions which also required changing from @{} to ${} for expression evaluation.


packages/dialog/templates/en-us/stringEntity.lg.lg
```diff
@@ -1,10 +1,10 @@
 # filename
-- @{entityLGFilename()}
+- ${entityLGFilename()}
 
 >>> This is because of issues around ' and } in strings--see enumEntityDesired.lg for a test case
 # template
 -```
- # @{property}Entity(val)
+ # ${property}Entity(val)
  - @\{val\}
```
ninggao: $\\{val\\}

packages/dialog/templates/en-us/enumEntity.lg.lg
```diff
@@ -1,39 +1,39 @@
...
  [library.lg]
  [library-Choose.lg]
 ```
 
 # chooseEntity
 -```
- # choose@{property}Entity
- - @\{chooseEnumEntity(@{concat("'", property, "'")})\}
+ # choose${property}Entity
+ - @\{chooseEnumEntity(${concat("'", property, "'")})\}
```
ninggao: It seems that a lot of the @\\{ cases are not replaced.

packages/dialog/templates/en-us/enumEntity.lg.lg
```diff
@@ -1,39 +1,39 @@
...
 
 # propertyEntity
-- @{join(createArray(namePlusSwitch(), cases(), default()), '')}
+- ${join(createArray(namePlusSwitch(), cases(), default()), '')}
 
 # namePlusSwitch
 -```
- # @{property}Entity(value) 
+ # ${property}Entity(value) 
  - SWITCH: @\{value\}
```
ninggao: replace with $

packages/dialog/templates/en-us/numberEntity.lg.lg
```diff
@@ -1,10 +1,10 @@
 # filename
-- @{entityLGFilename()}
+- ${entityLGFilename()}
 
 >>> This is because of issues around ' and } in strings--see enumEntityDesired.lg for a test case
 # template
 -```
- # @{property}Entity(val)
+ # ${property}Entity(val)
  - @\{val\}
```
ninggao: $\\{val\}

PR #610, opened 2020-02-26
Make build tests verbose

mdrichardson:
Currently, during a build where the tests fail, the console output isn't terribly helpful. It will look something like this:

image

...and has no indication of which tests failed.

By adding `-v` to `rush test`, this will show success/failure of every test as well as the error message.


mdrichardson:
@munozemilio Any idea why the build is giving this error? 

> ##[error]Could not find token for autheniticating. Please enable OAuth token in Build/Release Options.

Do I need special permissions for a PR to trigger a valid pipeline run?

PR #621, opened 2020-02-29
Switch to interface/implements in dialog:merge.

chrimc62:


packages/dialog/src/library/component.schema
```diff
@@ -104,6 +104,10 @@
             "type": "boolean",
             "default": false
         },
+        "writeOnly": {
```
christopheranderson: Is this intentional?

PR #626, opened 2020-03-03
bf-lu: update some function to support load file content directly

lei9444:
Changes in this PR:
1. Optimize the interface of crossTrainer.crossTrain function to remove the dependency on LuObject and LuOptions class
2. Optimize cross-train.train function to make it accept both config path and config content which is useful for auto-detecting config in cross train CLI package
3. Add some help functions in fileHelper.ts to help above #1 and #2
4. adjust test cases to test non file path id


feich-ms:
@munozemilio, this PR is refactoring some function interfaces of crossTrainer based on composer feedbacks. Please let me know if you have any concerns.

PR #641, opened 2020-03-06
Fix bug in luis:build and support operations in generator

chrimc62:
This is in preparation to move this to experimental and out of CLI.
Fix a bug in luis:build where --dialog generates $type instead of $kind.
Redo how the generator works so that less is in the code and more in the templates.  In particular templates can generate the default $entities and include other templates.
Added templates to support manipulating enum arrays including add/remove/clear/show.
Added hash codes to all generated files.


.vscode/launch.json
```diff
@@ -148,6 +176,7 @@
             "name": "LUIS Build DynamicLists",
             "preLaunchTask": "${defaultBuildTask}",
             "program": "${workspaceFolder}/packages/lu/bin/run",
+            "outputCapture": "std",
```
coldplaying42, line 179: luis/bin/run

.vscode/launch.json
```diff
@@ -148,6 +176,7 @@
             "name": "LUIS Build DynamicLists",
             "preLaunchTask": "${defaultBuildTask}",
             "program": "${workspaceFolder}/packages/lu/bin/run",
```
coldplaying42, line 178: should be luis/bin/run

coldplaying42, line 178: should be luis/bin/run

.vscode/launch.json
```diff
@@ -148,6 +176,7 @@
...
+            "outputCapture": "std",
             "outFiles": [
                 "./packages/lu/lib/**"
```
coldplaying42, line 181: packages/lu should be changed to packages/luis

packages/dialog/src/commands/dialog/readme.md
```diff
@@ -35,14 +35,15 @@ The generate command generates .lu, .lg, .qna and .dialog assets from a schema d
...
 
 Globally there are a few extra keywords including:
 - **\$public** List of the public properties in the schema. By default these are the top-level properties in the root schema and this will be automatically generated if missing.
 - **\$requires** A list of JSON Schema to include.  This is different than $ref in that you can either use a URL or just a filename which will be looked for in the template directories.  If you want to include the standard confirmation/cancel/navigation functionality you should include include `standard.schema`.
 - **\$expectedOnly** A global list of entities which will only be bound to properties if the property is expected.  This can be overriden for specific properties as well.
 - **\$triggerIntent** Name of the trigger intent or by default the name of the schema.
 - **\$templates** Global templates to include.
+- **\$defaultOperation** Default operation to use for assigning entities to properties.
```
coldplaying42, line 46: Do we need to add the definition of $operation here?

packages/dialog/src/library/dialogGenerator.ts
```diff
@@ -21,106 +24,126 @@ export enum FeedbackType {
...
+        let json = JSON.parse(file)
+        let oldHash = json.$Generator
+        if (oldHash) {
+            delete json.$Generator
+            let hash = computeJSONHash(json)
+            result = oldHash === hash
+        }
+    }
+    return result
+}
```
coldplaying42, line 79: When we modify lg, lu, dialog files by Composer, should we update the file hash at the same time?  In that case, the isUnchanged function always returns true.

packages/dialog/src/commands/dialog/readme.md
chrimc62: I still need to update the documentation, will do under another check-in.

---
In reply to: [402882721](https://github.com/microsoft/botframework-cli/pull/641#discussion_r402882721) [](ancestors = 402882721)

packages/dialog/src/library/dialogGenerator.ts
chrimc62: No, the idea is we want to know if composer has modified it--that way if it is unmodified we know we can replace it.

---
In reply to: [402887456](https://github.com/microsoft/botframework-cli/pull/641#discussion_r402887456) [](ancestors = 402887456)

common/scripts/install-run.js
```diff
@@ -58,45 +57,56 @@ function _parsePackageSpecifier(rawPackageSpecifier) {
  * we'd prefer to skip that line and continue looking in other places such as the user's
  * home directory.
  *
+ * IMPORTANT: THIS CODE SHOULD BE KEPT UP TO DATE WITH Utilities._copyNpmrcFile()
```
munozemilio, line 60: Why this change?

packages/luis/test/fixtures/luisgen/obj/Debug/netcoreapp2.1/LUISGenTest.AssemblyInfoInputs.cache
```diff
@@ -0,0 +1 @@
+c9fa663dac6390a7eefc64231f367790aa287411
```
munozemilio, line 1: Should this be committed?

chrimc62: It keeps getting recreated--I'll add it to .gitignore instead--it should already be excluded because of obj\debug.

---
In reply to: [403170119](https://github.com/microsoft/botframework-cli/pull/641#discussion_r403170119) [](ancestors = 403170119)

rush.json
```diff
@@ -1,7 +1,7 @@
 {
   "$schema": "https://developer.microsoft.com/json-schemas/rush/v5/rush.schema.json",
-  "rushVersion": "5.19.1",
-  "pnpmVersion": "4.8.0",
+  "rushVersion": "5.20.0",
+  "pnpmVersion": "4.9.3",
```
chrimc62, line 4: I rolled this back.

PR #654, opened 2020-03-11
qnamaker build CLI

feich-ms:
QnA build CLI implementation for https://github.com/microsoft/botframework-cli/blob/master/specs/qnamaker-build.md.

To test, go to packages/qnamaker/bin folder and run below command: 
run qnamaker:build --in ../test/fixtures/testcases/qnabuild/sandwich/qnafiles/sandwich.en-us.qna --subscriptionKey xxx --botName xxx --dialog --log

Update 4.7:
Multiturn is supported now. Multiturn with references is also supported. Corresponding tests are also added.

@vishwacsena @boydc2014, you can have a try to see if there are any issues.


packages/lu/src/parser/utils/serviceBase.js
```diff
@@ -0,0 +1,83 @@
+/**
```
munozemilio, line 1: Since this is only for qnamaker build, should we rename it and move it under qnabuild?

feich-ms: fixed. Moved this file to qnabuild folder.

feich-ms:
@vishwacsena, I have reverted the file and url reference related logic in lu build core functionality. Meanwhile, I exposed two new functions naming **importFileReference** and **importUrlReference**. They import url or file reference into kb and return the extracted kb pairs with qna format. You can refer the qnabuild.test.js to see how to call them. Of course, they can be adjusted based on composer requirements in the future.

vishwacsena:
@feich-ms this is looking good. I pushed up minor changes via [this commit][1]

[1]:https://github.com/microsoft/botframework-cli/pull/654/commits/ba732038157270c32c792e6ef37e357eab86aefe

vishwacsena:
@munozemilio can you take a look and review the updates based on your PR feedback? Thanks!

feich-ms:
@vishwacsena, I have updated the code/tests/doc per our discussion. All the qna files with same locale(default to en-us) will be merged into one and then published. The kb name will be **botName(suffix).locale.qna**. The dialog assets will also be named with **botName.locale.qna.dialog**(Recognizer) or **botName.qna.dialog**(MultiRecognizer). Please verify. If no other issues, we can merge this one.

Issue #653, opened 2020-03-11
luis:build treats = as part of name


chrimc62:
If you have a file like:
@ list foo=
   - woof:

Then the name of the entity will be considered to be 'foo='.  It should just be 'foo'.
[bug]

PR #657, opened 2020-03-11
LU fixes

### LINKED ISSUE #652: luis:build turns all labels into ml even if defined as composite in another file
## Describe the bug
File1: 
@ list blah=
- something:

@ composite foo = [blah]

File2:
# utterances
- @{foo= something}

File3: 
Include both File1 and File2

Try to do luis:build on file3.

You will get an error: Entity names must be unique. Duplicate definition found for "foo". Prior definition - '@ ml foo'

The reason is that the collator on file 2 ends up assuming foo is an ml entity which conflicts with its definition as a composite.

## Expected behavior
You can distribute definitions including labeling. 

[bug]

vishwacsena:
Fixes #652 #653

### DIFF
--- a/packages/lu/src/parser/lufile/parseFileContents.js
+++ b/packages/lu/src/parser/lufile/parseFileContents.js
@@ -1034,7 +1034,7 @@ const parseAndHandleEntityV2 = function (parsedContent, luResource, log, locale)
         for (const entity of entities) {
             if (entity.Type !== INTENTTYPE) {
                 entity.Name = entity.Name.replace(/^[\'\"]|[\'\"]$/g, "");
-                let entityName = entity.Name;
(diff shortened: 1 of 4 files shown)


packages/lu/src/parser/lufile/parseFileContents.js
```diff
@@ -1034,7 +1034,7 @@ const parseAndHandleEntityV2 = function (parsedContent, luResource, log, locale)
         for (const entity of entities) {
             if (entity.Type !== INTENTTYPE) {
                 entity.Name = entity.Name.replace(/^[\'\"]|[\'\"]$/g, "");
-                let entityName = entity.Name;
+                let entityName = entity.Name.endsWith('=') ? entity.Name.slice(0, entity.Name.length - 1) : entity.Name;
```
chrimc62: >entity.Name.endsWith('=') [](start = 33, length = 25)

This will work, but isn't there a limited set of chars allowed in entity names for LUIS?

PR #671, opened 2020-03-16
expose getActiveVersionIds function in lubuild

### LINKED ISSUE #669: bf-lu: get the applications' version after publishing
Now use bf-lu as a library to publish the lu file will return nothing. In composer we want to show the published version in all up view. Can we add a return value to get the published applications' version?

feich-ms:
expose getActiveVersionIds function for composer that is going to use this to detect online app version. fix #669

### DIFF
--- a/packages/lu/src/parser/lubuild/builder.ts
+++ b/packages/lu/src/parser/lubuild/builder.ts
@@ -253,6 +253,32 @@ export class Builder {
     return writeDone
   }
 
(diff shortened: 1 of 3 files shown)


feich-ms:
Sure, will add.

feich-ms:
unit tests added

feich-ms:
@lei9444 will this meet your requiements?

PR #672, opened 2020-03-16
Fix comparing luis applications bug in lu build

feich-ms:
fix a bug in lu build when comparing apps to decide update app or not. The bug is caused due to the returning application structure of luis exportApplication api changes in version 3.0-preview that I switched recently. The reason why unit tests cannot hit this case is that the returning json from luis portal doesn't change in new API version but the object from luis api changed, so it's hard to detect it by unit tests. **_Hope this can catch R8 release_**, otherwise lu build may hit issues when users run lu build for lu files which don't have changes(That's saying lu build will always update luis application to a new version even though the lu files don't change actually). @vishwasena, @munozemilio could you help?

### DIFF
--- a/packages/lu/src/parser/lubuild/core.ts
+++ b/packages/lu/src/parser/lubuild/core.ts
@@ -87,6 +87,15 @@ export class LuBuildCore {
 
     currentApp.name = existingApp.name
 
+    existingApp.model_features = existingApp.phraselists
+    delete existingApp.phraselists
+
+    existingApp.regex_features = existingApp.regex_features
+    delete existingApp.regex_features
+
+    existingApp.regex_entities = existingApp.regex_entities
+    delete existingApp.regex_entities
+
     return !this.isApplicationEqual(currentApp, existingApp)
   }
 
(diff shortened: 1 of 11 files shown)


vishwacsena:
Not a P0 since the real impact is reimport of application version when there are no changes TO LU file.

Can you please add tests?

feich-ms:
OK. It's hard to add unit tests since the structure of returning application from luis api often changes. I think the best way to solve this issue is removing emtpy properties from both lu converted app and api got app since it's not easy to always keep our luis application definition to align with the one from luis API. I will try to fix in this way and add unit tests

vishwacsena:
Could you try this rather than lubuild having to maintain this logic?

1. User input LU - > parse to JSON - > To LU
2. Exported JSON - > To LU

compare 1 and 2 after you convert both to lower case.?

Might be lesser code in your side. We need to keep #1 and #2 up to date with LUIS APIs in BF-LU any ways.

feich-ms:
Good suggestion. Will have a try.

feich-ms:
@vishwacsena, I spent much time today on this bug, but made little progress. The json returned from luis api has many unexpected properties or behaviors. e.g. 1. It's phraselists corresponds to model_features in our lu format and it added enabledForAllModels: true automatically in phraselists. 2. It automatically adds useFeaure definition below each intent if there is a phrase list definton in the lu file even though I didn't define anything below those intents. These corner cases make the convertion of luis api json to lu content not align with the one converted from our input. Do you have more suggestions?

vishwacsena:
@feich-ms, we should just fix lu parser so it sets the same defaults on LU to JSON. Can you list the specific things that are set by LUIS automatically as a separate ticket? Let's fix that so you can just compare LU and stay agnostic to the actual JSON representation.

feich-ms:
@vishwacsena, I created a issue here https://github.com/microsoft/botframework-cli/issues/675 to track the schema disalignments or unexpected behaviors between BF-LU and Luis API. For this bug, I adjusted the schemas and skipped unexpected behaviors when comparing lu content. It can work well now. But the comparison functionality is still weak if we don't really align the schemas and behaviors between BF-LU and LUIS API.

Issue #676, opened 2020-03-19
Lu parser takes too long when paring simple intent section that has a large number of entity definitions


feich-ms:
## Describe the bug
Given a lu file which has a intent defintion and a large number of entity definition under that intent, Any CLI tool that is calling lu parser will take a long time to parse the content with antlr. This is a bug caused by the rules in LUFileParser.g4. EntitySection and NewEntitySection are both root sections of paragraph and child sections of simpleIntentSection which may cause some unexpected loop.

## To Reproduce
1. create a lu file with a large number of entity definitions under a intent
2. run 'bf luis:convert --in {lu file path}'
3. See the execution time of this command

## Expected behavior
It should take reasonable time to parse lu content

## Additional context

[bug]

feich-ms:
fixed by PR #677

PR #677, opened 2020-03-19
Optimize LUFileParser.g4 to accelerate parsing of simple intent section with a large number of entities

feich-ms:
fix issue #676. Before the fix, the lu parser takes unreasonable time to parse lu content with a large number of entity definitions under an intent. e.g. 100 entities will cost 4000+ milliseconds. After the fix, it takes 200-300 milliseconds.


packages/lu/src/parser/lufile/LUFileParser.g4
```diff
@@ -49,7 +47,7 @@ subIntentDefinition
     ;
 
 simpleIntentSection
-    : intentDefinition (entitySection | newEntitySection)*
+    : (intentDefinition? (entitySection | newEntitySection)+) | intentDefinition
```
boydc2014: Can you explain why this would fix this issue?

boydc2014:
Please attach with root cause analysis and the explanation of the fix

feich-ms:
**Root cause:** entitySection(and newEntitySection) are child rules of both paragraph and simpleIntentSection, and simpleIntentSection is child rule of paragraph, that means each entitySection following an intent will first be identified as entitySection in paragraph and then look back to search bigger scope to match simpleIntentSection. When there are a large number of entities, this knid of back search will be very expensive.

**Why this fix works:** To fix this, we can remove entitySection(and newEntitySection) from either paragraph or simpleIntentSection to avoid this kind of back search. We choose removing it from paragraph. But to keep our logic that entitySection can be standalone, we optimize the simpleIntentSection rule to make it able to match standalone entitySection. Below simpleIntentSection rule will match standalone intent, standalone entity, intent with entities. And then we adjusted the code in luParser.js to consume the new standalone entity context. This fix solve our issue and reduce the paring time signally.
```
simpleIntentSection
    : (intentDefinition? (entitySection | newEntitySection)+) | intentDefinition
    ;
```

packages/lu/src/parser/lufile/LUFileParser.g4
feich-ms: explained below

feich-ms:
@vishwacsena @munozemilio could you approve this PR?

feich-ms:
@vishwacsena @munozemilio I merged this minor fix of parser rules, please let me know if you have concerns.

PR #684, opened 2020-03-23
fix: bf-lu bug fixes 

### LINKED ISSUE #675: Align Luis json schema between BF-LU and LUIS API
The luis schema converted from lu content in bf-lu does not align with the one returned from Luis API which makes the comparison of local luis model and remote luis model very difficult. Lu build CLI depends on the comparison result to decide whether to repunlish the luis applicationto a new version or not. 

Align the json schema between BF-LU and LUIS API

**Describe alternatives you've considered**
1. Change the Luis schema in bf-lu to align with the schema from API
2. Make some changes in luConverter.js to make sure luisToLuContent function can convert to the same lu content for local and remote luis json that from same original lu content

**Additional context**
I have found several disalignments such as

1. Schema disalignments: BF-LU <-> LUIS API
    - model_features <-> phraselists
    - regex_entities <-> regexEntities
    - regex_features <-> regexFeatures
    - the luis_schema_version returned from LUIS API is always 6.0.0 even though I set it to 3.2.0 when importing
2. Unexpected behaviors
    - The json returned from luis API always automatically add features defintion to each intent if there is a phraseList definition in lu file but not enabled for each intent
    - The json returned from luis API sometimes automatically set settings with NormalizeDiacritics: true even though nothing is set for this when importing
    - The json returned from luis API automatically set enabledForAllModels to be true even though it is not set when importing

To repro this issue, you can publish a lu file with bf luis:build and get the json from API with bf:version:export. You will find above disalignments.

[enhancement]

vishwacsena:
Fixes #675 

- pattern.any entities if present in LUIS application JSON will now be represented in LU format as well
- Automatically add phrase list as feature to intent and entities (if the phraselist is not specifically added as a feature to specific models)
- Automatically set enabledForAllModels for phraseList.


packages/lu/test/parser/lufile/parseFileContents.parseFile.test.js
```diff
@@ -261,7 +261,7 @@ m&m,mars,mints,spearmings,payday,jelly,kit kat,kitkat,twix
                         .then(res => {
                                 assert.equal(res.LUISJsonStructure.entities.length, 1);
                                 assert.equal(res.LUISJsonStructure.entities[0].name, 'product');
-                                assert.equal(res.LUISJsonStructure.model_features.length, 1);
```
feich-ms: model_features will be replaced by phraseLists in bf-lu luis json, right? If so, I will remove the exception handling for model_features in lubuild core.

vishwacsena: Not entirely. This PR just makes sure that phraselists are automatically added as features to intents. So if you are using the new entity or feature assignment (LUIS schema version 6.0.0), you will no longer see model_features.

vishwacsena:
@feich-ms 
- service exports `regex_entities`. can you clarify where you see `regexEntities`? 
- regex_features are not supported in LU format. Not sure where they are supported even in the portal experience. AFAIK this is legacy feature.

It would be good for you to evaluate/ test this PR for your comparison needs. I want BF-LU to abstract out transform issues and not have this duplicate this everywhere. Can you try and consume this fix, and compare a) LU -> JSON -> in memory LU b) LUIS JSON -> LU and compare them to decide if we should update app version?

feich-ms:
@vishwacsena 

- for regex_entities, luis api returns regexEntities actually. You can publish a lu file with regex entity definitions with lu build and then use luis:version:export the get the luis json file, you will find it is regexEntities instead of regex_entities. Please use bf luis:version:export command instead of downloading it from portal.
- for regex_features, since it is not used in lu file, I will ignore it and leave as it is.

feich-ms:
@vishwacsena I pushed a commit here https://github.com/microsoft/botframework-cli/pull/684/commits/1ef2e0467654079b3a9f961630e76d28bd043be6. In this commit, I did below changes:

- removed schema conversions between regex_entities vs regexEntities, model_features  vs phraselists, regex_entities vs regexEntities in lu build side
- move luis properties sort functionality to luis.js.

After above changes, the comparison of luis josn in lubuild side will be pure like a) LU -> JSON -> in memory LU to compare with b) LUIS JSON -> LU

All other logic are moved to lu side:

- handle regexEntity exception in luis.js
- add sort functionality in luis.js to make sure parseToLuContent return same content for luis jsons that just property orders are different

At last, I added back the phrase list defintions in lu build tests, it works well now. phraselists defintions in lu file are added to each intents automatically just like luis api does.

packages/lu/src/parser/luis/luis.js
```diff
@@ -48,6 +55,11 @@ const initialize = function(instance, LuisJSON) {
     for (let prop in instance) {
         instance[prop] = LuisJSON[prop];
     } 
+
+    // add regexEntities property that returned from luis export api
+    const regexEntities = 'regexEntities';
```
vishwacsena: Unclear why this is required. Can you clarify? If LUIS JSON had the property, then lines 48-50 will assign them to instance. Why special case regex entities?

feich-ms: No exactly, line 48-50 is looping property defined in luis(Line 8-18) that do not include regexEntities property

feich-ms: We can also add regexEntity in luis schema then we can remove the code piece here.

packages/lu/src/parser/lufile/parseFileContents.js
```diff
@@ -183,10 +183,38 @@ const parseLuAndQnaWithAntlr = async function (parsedContent, fileContent, log,
...
+    updateIntentAndEntityFeatures(parsedContent.LUISJsonStructure);
+
     helpers.checkAndUpdateVersion(parsedContent.LUISJsonStructure);
 
 }
 
+const updateIntentAndEntityFeatures = function(luisObj) {
+    let plForAll = luisObj.model_features.filter(item => item.enabledForAllModels == undefined || item.enabledForAllModels == true);
+    plForAll.forEach(pl => {
+        luisObj.intents.forEach(intent => {
```
munozemilio: Can this be abstracted to a function to avoid code duplication?

packages/lu/src/parser/luis/luis.js
```diff
@@ -40,6 +40,13 @@ class Luis {
     validate() {
         return validator(this)
     }
+
+    sort() {
```
munozemilio: Initially we decided to remove the sort functionality from the classes and leave this to be implemented by te consumer. I understand this is used by luis build but I do prefer to keep this hidden. In this case Luis Build can have a sorting function which receives the luis class. My intention is to keep everything decoupled since we may split lubuild and lu library.

feich-ms: Make sense. My intention to put this sort function here is to make sure all the disalignments for luis converted lu comparisons can be handled here instead of in two places. Anyway, I can move this to lu build.

feich-ms: resolved

vishwacsena:
@munozemilio fixed PR feedback! can you take a look? Thanks!

PR #688, opened 2020-03-25
refactor evaluation API

hcyang:


packages/dispatcher/src/mathematics/mathematics_helper/MathematicsHelper.ts
```diff
@@ -2190,6 +2190,321 @@ export class MathematicsHelper {
         return [sparseIndexArray0, denseValueArray0];
     }
 
+    public static vectorSparseMapDenseArrayAssignRandomTo(
+        sparseIndexArray0: number[],
+        sparseValueArray0: number[],
+        denseValueArray0: number[]): [number[], number[]] {
+        for (const index of sparseIndexArray0) {
+            denseValueArray0[index] = Utility.getRandomNumber();
+        }
```
daveta: Might want to switch to "for" instead of "for-of".
[https://hackernoon.com/3-javascript-performance-mistakes-you-should-stop-doing-ebf84b9de951](https://hackernoon.com/3-javascript-performance-mistakes-you-should-stop-doing-ebf84b9de951)

hcyang: I was trying to avoid a style warning, thus used 'for-of' instead. Would hope Typescript/Javascript can improve 'for-of' as there is no reason not to.

Expected a 'for-of' loop instead of a 'for' loop with this simple iteration (prefer-for-of)tslint(1)

        for (let i = 0; i < sparseIndexArray0.length; i++){
            denseValueArray0[sparseIndexArray0[i]] = Utility.getRandomNumber();
        }

packages/dispatcher/package.json
```diff
@@ -9,7 +9,8 @@
     "@microsoft/bf-lu": "1.0.0",
     "@oclif/command": "~1.5.19",
     "@oclif/config": "~1.13.3",
-    "tslib": "^1.10.0"
+    "tslib": "^1.10.0",
```
tsuwandy: @hcyang , shouldn't this be dev dependencies?

PR #690, opened 2020-03-26
cross train CLI

### LINKED ISSUE #687: cross-train -> master
Filing this ticket to get cross-train into master branch. 

Need detailed docs.

feich-ms:
Fixes #687. The usage is listed here https://github.com/microsoft/botframework-cli/blob/cross-train/packages/cross-train/README.md


vishwacsena:
@feich-ms for this one can you confirm that if `--in` is to a folder, that you recursively find all .qna files and collate to a single KB per lang x locale?

feich-ms:
@vishwacsena, I think you are saying to merge all lu files of same locale into one file when publishing. This should be done in qna build, not this cross train branch I think, right?

vishwacsena:
Yes. sorry. left the comment in the wrong PR. But yes, I was referring to qnamaker:build PR :)

vishwacsena:
@feich-ms I was playing with this today and I believe we should re-target this as two separate commands one under luis and another under qnamaker. This way instead of a single top level command, the luis:cross-train will sit as a peer to luis:build and can cross train across LUIS apps as well as with qnamaker. In a similar way we should have qnamaker:cross-train sit as a peer to qnamaker:build and cross-train across qna KBs as well as LU files. The sequence of commands the user then runs will be this - 
```
bf luis:cross-train
bf luis:build --dialog --cross-train
bf qnamaker:cross-train
bf qnamaker:build --dialog --cross-train
```

We also need to take a related change to luis:build and qnamaker:build to write out cross-trained recognizer configuration when `--dialog` and `--cross-train` is specified. Note: given the succession of commands above, we need to first examine to see if a dialog file exists w/ a cross-trained recognizer and then update it (e.g. user first added luis and then qna etc). Also if user did not run one of the luis or qnamaker:build commands without the --cross-train and --dialog flag, we should throw and error. 

Can you please update this PR to 
1. Remove cross-train from top level command
2. Add cross-train commands to both luis and qnamaker
3. Add --cross-train to luis and qnamaker build so we spit out a crossTrainedRecognizer configuration? Without this flag, --dialog will write out a regular `LuisRecognizer` or `QnAMakerRecognizer` configuration. 

@boydc2014 as FYI.

feich-ms:
@vishwacsena, I have some questions
1. I'm not 100% sure I understand you meanings of below words. Can you explain more? `we need to first examine to see if a dialog file exists w/ a cross-trained recognizer and then update it (e.g. user first added luis and then qna etc).`
2. Similarily, I'm not 100% understand below words. `Also if user did not run one of the luis or qnamaker:build commands without the --cross-train and --dialog flag, we should throw and error.`
3. What is the format of **crossTrainedRecognizer** configuration?

vishwacsena:
```
I'm not 100% sure I understand you meanings of below words. Can you explain more? we need to first examine to see if a dialog file exists w/ a cross-trained recognizer and then update it (e.g. user first added luis and then qna etc).
```
So when user runs bf `luis:cross-train` followed by `bf luis:build --dialog --cross-train`, we will examine the .dialog file for each dialog to see if it has a `luisrecognizer` configuration or a `crosstrainedrecognizerset` configuration. If it is a `luisrecognizer` configuration, then we convert it into a `crosstrainedrecognizerset` configuration. Running `luis:build` without the `--cross-train` should revert the configuration to just `luisrecognizer`. Next when user runs `qnamaker:cross-train` followed by `qnamaker:build --dialog --cross-train`, we will see that we already have a `crosstrainedrecognizerset` configuration and so will simply just add the .qna to the crosstrained recognizer set. If we do not see a `crosstrainedrecognizerset` then we should upgrade the recognizer to a cross-trained one but preserve the prior luis recognizer configuration. Does that help? 

```
Similarily, I'm not 100% understand below words. Also if user did not run one of the luis or qnamaker:build commands without the --cross-train and --dialog flag, we should throw and error.
```

Upgrade from an individual luis/qna -> crosstrained if user specifies `--cross-train` option. Now if user misses `--cross-train` but we detect a crosstrainedrecognizerset configuration, then we should throw an error because a dialog can at most have only one recognizer and so we would not know which one to keep. 

```
{
     "$kind": "Microsoft.CrossTrainedRecognizerSet",
     "recognizers": [
           "travelbot.lu",
           "travelbot.qna"
     ]
}
```

I think this is right. I have a sample [here][1]

[1]:https://github.com/microsoft/botbuilder-dotnet/blob/b75a1755b7ddfe4e3a4ec136bd9c68b8429a628d/samples/TestBed/Dialogs/RootDialog%20-%20QnA/RootDialogQnA.cs#L376

feich-ms:
@vishwacsena, I pushed some commits today. It can convert luis/qnamaker recognizer to crosstrained recogizer. But I have a question here. What is the file name for the crosstrained recognizer? Currently, for luis build or qna build dialogs, we have below naming solution:
1. luis recognizer: fileName.en-us.**lu**.dialog
2. luis multi recognizer: fileName.**lu**.dialog
3. qna recognizer: botName.en-us.**qna**.dialog
4. qna multi recognizer: botName.**qna**.dialog

In my commits, I'm using the same name for luis recognizer and luis cross trained recognizer, which means I'm converting luis recognizer to crosstrained recognizer in the same file named fileName.en-us.**lu**.dialog. In this case, I can not merge qna crosstrained recognizer into luis crosstrained recognizer file since they have different file names. So the question is still this: which file name for crosstrained recognizer?

vishwacsena:
@feich-ms starting with the simple use case, if i have dialogA and a dialogA.lu and dialogA.qna, then I would expect to find the crossTrained recognizer in `dialogA.lu.qna.dialog`. Having both `.lu.qna` would signify crossTrained. We could then extrapolate the convention and say `dialogA.lu.dialog` will be the multiLingual recognizer and `dialogA.<locale>.lu.dialog` be the individual luisRecognizer configuration. 

This would also mean that if we detect `dialogA.qna` for `dialogA` then we should write out `dialogA.qna.dialog` with the multiLanguageRecognizer and `dialogA.<locale>.qna.dialog` as the qnamaker recognizer and have `dialogA.lu.qna.dialog` be the cross trained recognizer configuration. 

Let's also simplify and prune the --dialog options to be `--dialog = multiLanguage [default] | crossTrained`. We can always add `--dialog Luis | qnamaker` for R10 or later.

feich-ms:
@vishwacsena, I pushed more commits. I don't know whether current implementations meet your requirement. But I can show how to use it and you can have a try and then give feedbacks.
1. bf luis:cross-train or bf qnamaker:cross-train --in xxx --out xxx
2. bf luis:build --in xxx --authoringKey xxx --botName xxx --out xxx --dialog crosstrained
3. bf qnamaker:build --in xxx --subscriptionKey xxx --botName YYY --out xxx --dialog crosstrained

Please notice that botName in qnamaker build should be specified the same name of it's lu file (not botName in luis:build cmd) because we agreed that qnamaker should merge all qna files of same locale and publish with the name of botName.<locale>.qna.

I'm not confident I 100% got your ideas of how this should work. Anyway, please let me know any feebacks at any time.

cleemullins:
IS this ready to merge?

munozemilio:
@cleemullins I'll do a second pass since a lot of changes were added since my last review.
I will merge it after that

packages/lu/src/parser/cross-train/confighelper.js
```diff
@@ -0,0 +1,107 @@
...
+
+  if (!fs.existsSync(rootLuPath)) {
+    throw (new exception(retCode.errorCode.INVALID_INPUT, `Failed to parse mapping rules config from file system: ${rootLuPath} does not exist. Please provide config file by --config`))
+  }
+
+  rootDialog.triggers.forEach(trigger => {
+    if (trigger.$type && trigger.$type === 'Microsoft.OnIntent') {
+      const actions = trigger.actions || []
+      for (const action of actions) {
+        if (action.$type && action.$type === 'Microsoft.BeginDialog') {
```
munozemilio, line 77: action.$type && action.$type !== 'Microsoft.BeginDialog' continue

packages/lu/src/parser/cross-train/confighelper.js
```diff
@@ -0,0 +1,107 @@
...
+          const dialogName = action.dialog
+          const target = dialogs.find(dialog => path.basename(dialog.path, dialogExt) === dialogName)
+          if (target) {
```
munozemilio, line 80: same here if !target continue

packages/lu/src/parser/qnabuild/builder.ts
```diff
@@ -285,23 +287,23 @@ export class Builder {
...
         if (force || !fs.existsSync(outFilePath)) {
           this.handler(`Writing to ${outFilePath}\n`)
-          await fs.writeFile(outFilePath, content.content, 'utf-8')
+          await this.writeDialog(content.content, outFilePath, dialogType, files)
           writeDone = true
         }
       }
     } else {
       for (const content of contents) {
         if (force || !fs.existsSync(content.path)) {
```
munozemilio, line 304: Code duplication...Can this be reduced? From 296 to 299 is the same as 304 to 307

packages/luis/src/commands/luis/cross-train.ts
```diff
@@ -0,0 +1,50 @@
...
+    rootDialog: flags.string({description: 'rootDialog file path which is relative to folder specified by --in. If --config not specified, cross-trian will automatically construct the config from file system based on root dialog file'})
+  }
+
+  async run() {
+    try {
+      const {flags} = this.parse(LuisCrossTrain)
+
+      if (!path.isAbsolute(flags.in)) flags.in = path.resolve(flags.in)
+      if (flags.config && flags.config !== '') {
+        if (!path.isAbsolute(flags.config)) flags.config = path.join(flags.in, flags.config)
```
munozemilio, line 29: && condition?

packages/luis/src/commands/luis/cross-train.ts
```diff
@@ -0,0 +1,50 @@
...
+      } else if (flags.rootDialog && flags.rootDialog !== '') {
+        if (!path.isAbsolute(flags.rootDialog)) flags.rootDialog = path.join(flags.in, flags.rootDialog)
```
munozemilio, line 31: && condition?

packages/qnamaker/src/commands/qnamaker/cross-train.ts
```diff
@@ -0,0 +1,50 @@
...
+    rootDialog: flags.string({description: 'rootDialog file path which is relative to folder specified by --in. If --config not specified, cross-trian will automatically construct the config from file system based on root dialog file'})
+  }
+
+  async run() {
+    try {
+      const {flags} = this.parse(QnamakerCrossTrain)
+
+      if (!path.isAbsolute(flags.in)) flags.in = path.resolve(flags.in)
+      if (flags.config && flags.config !== '') {
+        if (!path.isAbsolute(flags.config)) flags.config = path.join(flags.in, flags.config)
```
munozemilio, line 29: && condition?

packages/qnamaker/src/commands/qnamaker/cross-train.ts
```diff
@@ -0,0 +1,50 @@
...
+    config: flags.string({description: 'path to config file of mapping rules which is relative to folder specified by --in. If not specified, it will read default config.json from the folder specified by --in'}),
+    intentName: flags.string({description: 'Interruption intent name', default: '_Interruption'}),
+    rootDialog: flags.string({description: 'rootDialog file path which is relative to folder specified by --in. If --config not specified, cross-trian will automatically construct the config from file system based on root dialog file'})
+  }
+
+  async run() {
+    try {
+      const {flags} = this.parse(QnamakerCrossTrain)
+
+      if (!path.isAbsolute(flags.in)) flags.in = path.resolve(flags.in)
```
munozemilio, line 27: if possible add brackets. Its easier to read the code

munozemilio:
Minor nits added.

packages/lu/src/parser/cross-train/confighelper.js
feich-ms: fixed

packages/lu/src/parser/cross-train/confighelper.js
feich-ms: very make sense. fixed.

packages/lu/src/parser/qnabuild/builder.ts
feich-ms: good catch. fixed. Thx.

packages/luis/src/commands/luis/cross-train.ts
feich-ms: good catch. fixed. thx.

packages/luis/src/commands/luis/cross-train.ts
feich-ms: good catch. fixed. Thx.

packages/qnamaker/src/commands/qnamaker/cross-train.ts
feich-ms: fixed.

packages/qnamaker/src/commands/qnamaker/cross-train.ts
feich-ms: good suggestion. fixed. Thanks,

feich-ms:
@munozemilio Thanks for all your good comments. They all very make sense. I fixed and resolved them. So I will merge this PR. Please let me know if you have any concerns.

PR #691, opened 2020-03-30
fix bug when writing dialogs

feich-ms:
fix the bug when writing dialog assets. Before the fix, it will throw error if users didn't specify the --out parameter or specified a folder that didn't exist. After the fix, it will write to the folder of the cmd if --out is not specified and it will also create a foler if users specify a folder that does not exist.


feich-ms:
merge these small changes to another PR #704, so close this one.

Issue #673, opened 2020-03-17
qnamaker:kb:export (download) and qnamaker:kb:replace should support qnaformat


vishwacsena:
support export and replace to qnaformat (from service)

munozemilio:
Fixed by #695

PR #695, opened 2020-03-31
[qnamaker] Export as well as replace support for .qna format

### LINKED ISSUE #670: qnamaker:kb:export command not exporting multi turn prompts
## Versions
 @microsoft/botframework-cli/4.8.0-preview.100922 win32-x64 node-v12.14.1

v12.14.1

What OS are you using
Windows 10.1909 Build 18363.657

## Describe the bug
bf  qnamaker:kb:export command does not export multi-turn prompts.

## To Reproduce
1. Run command on any QnAMaker KB that includes multi-turn prompts
2. Check output json file and you will notice that prompts are missing for Q&A pairs that include them.

## Expected behavior
Expect that prompts will be exported

[bug]

vishwacsena:
Fixes #670 and #673

### DIFF
--- a/packages/qnamaker/src/utils/qnamakerbase.ts
+++ b/packages/qnamaker/src/utils/qnamakerbase.ts
@@ -9,6 +9,8 @@ const path = require('path')
 const qnaconfig = require('./../../utils/qnaconfig')
 const srvMan = require('./../../utils/servicemanifest')
 const {ServiceBase} = require('./../../utils/api/serviceBase')
+const file = require('@microsoft/bf-lu/lib/utils/filehelper')
+
 
 export async function processInputs(flags: any, payload: any, configfile: string, stdin = '') {
   let result: Inputs = {}
@@ -22,7 +24,12 @@ export async function processInputs(flags: any, payload: any, configfile: string
     flags.in = stdin ? stdin : flags.in
     result.requestBody = await srvMan.validateArguments(serviceManifest, flags)
     if (stdin || flags.in) {
-      result.requestBody = stdin ? JSON.parse(stdin) : await getFileInput(flags.in)
+      result.requestBody = stdin ? JSON.parse(stdin) : await file.getContentFromFile(flags.in)
+      try {
+        result.requestBody = JSON.parse(result.requestBody)
+      } catch(ex) {
+        flags.qnaFormat = true
+      }
     }
     config.endpoint = flags.endpoint
     result.config = config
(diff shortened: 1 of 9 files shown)


vishwacsena:
@feich-ms can you look at this PR in the context of qnamaker:build and comment if we can switch to use .qnaformat for both replace as well as export needs? I have not played with qnamaker:build yet but will do so after this is reviewed and merged in and you have been able to get multi-turn supported in qnamaker:build. Also I hope we do not do create w/ KB in qnamaker:build. It always has to be a create command followed by replace for qnamaker:build since the create command does not support .qna format.

packages/qnamaker/src/utils/qnamakerbase.ts
```diff
@@ -22,7 +24,12 @@ export async function processInputs(flags: any, payload: any, configfile: string
     flags.in = stdin ? stdin : flags.in
     result.requestBody = await srvMan.validateArguments(serviceManifest, flags)
     if (stdin || flags.in) {
-      result.requestBody = stdin ? JSON.parse(stdin) : await getFileInput(flags.in)
+      result.requestBody = stdin ? JSON.parse(stdin) : await file.getContentFromFile(flags.in)
```
munozemilio: I was thinking on this and maybe qnamaker commands can support piped .qna content. If that is the case then this should change.

feich-ms:
Will have a try

feich-ms:
@vishwacsena, I am OOF this Wednesday and Thursday, so I haven't got a chance to try it. But from tests, it seems to work well. One question is if I export with qna format, will it reconstruct the url and file references to qna files? If not, we can not compare the qna format contents to decide whether updating knowledge base just like we do in luis build. I will give updates soon when I'm back to work.

vishwacsena:
@feich-ms export will not reconstruct url and file references. The way to do it would be to complete cross-train etc on our side, put together the single qna KB (per locale) and compare it against the exported qna document. We might need to do some work on the comparison end to sort on questions and answers like we did for luis but other than that, that's how we should compare.

feich-ms:
@vishwacsena, I just tried to create an empty kb and then replace it with the multiturn qna content. It works now. But if I added a file or url referernce in the qna content, the referenced qna pairs are not included in published kb. So is it reasonable or valid if file or url references exist in multiturn qna file. What to do if file or url references are in multiturn qna content?

vishwacsena:
@feich-ms I believe we should only support URI reference. @vipeketi can you comment on how we should handle files? 
As for your proposal, immaterial of the KB content, why not just always do create followed by replace or just replace if KB already exists?

feich-ms:
@vishwacsena, I double tried and confirmed that URI and file reference are not supported under qna format. If they are not planned to support, I think we can always do step 3 to keep consistency for all lu files. If they are already suppoerted, we can drop the update api call for reference in step 2. Please help to confirm. Thanks.

vipeketi:
@feich-ms I find your approach of combining replace and update api as best approach for taking care of URI or file refernce updates.

PR #709, opened 2020-04-14
write out prediction options

feich-ms:
fix issue here https://github.com/microsoft/BotFramework-Composer/issues/2545

### DIFF
--- a/packages/lu/src/parser/lubuild/recognizer.ts
+++ b/packages/lu/src/parser/lubuild/recognizer.ts
@@ -39,10 +39,10 @@ export class Recognizer {
     this.endpointKey = '=settings.luis.endpointKey'
     this.predictionOptions = {
       includeAllIntents: '=coalesce(settings.luis.preditionOptions.includeAllIntents, false)',
-      includeInstanceData: '=coalesce(settings.luis.preditionOptions.includeInstanceData, false)',
-      log: '=coalesce(settings.luis.preditionOptions.log, false)',
-      preferExternalEntities: '=coalesce(settings.luis.preditionOptions.preferExternalEntities, false)',
-      slot: "=coalesce(settings.luis.preditionOptions.slot, '')",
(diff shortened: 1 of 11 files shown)


feich-ms:
added all the properties

packages/lu/src/parser/lubuild/recognizer.ts
```diff
@@ -36,6 +37,14 @@ export class Recognizer {
     this.applicationId = `=settings.luis.${targetFileName.split('.').join('_')}`
     this.endpoint = '=settings.luis.endpoint'
     this.endpointKey = '=settings.luis.endpointKey'
+    this.predictionOptions = {
+      includeAllIntents: '=coalesce(settings.luis.preditionOptions.includeAllIntents, false)',
+      includeInstanceData: '=coalesce(settings.luis.preditionOptions.includeInstanceData, false)',
+      log: '=coalesce(settings.luis.preditionOptions.log, false)',
```
luhan2017, line 43: Let's update the default value:

log = true
preferExternalEntities = true
slot = 'production'

luhan2017: @vishwacsena, we should add the default values in sdk schema. There are default values in sdk code but not in schema.

vishwacsena: @chrimc62 what should we default these values to?

chrimc62: includeAllIntents = false
includeInstanceData = true (Without this, entity events would not work.  You could make a case for this being false by default, but then your docs would have to tell people to turn it on--I'd rather it were true and someone that wants to make payloads smaller could turn it off.)
log = true
preferExternalEntities = true

feich-ms: changed to above default values. Please verify.

feich-ms:
Can anyone approve this PR if it looks good to you? @vishwacsena @luhan2017

PR #710, opened 2020-04-14
fix source .vs. ID order in JSON -> QnA format

vishwacsena:
Tien discovered this issue.Simple fix to get both the JSON -> QnA as well as QnA -> JSON be consistent in the order information should appear in .qna format.

### DIFF
--- a/packages/lu/src/parser/qna/qnamaker/qnaConverter.js
+++ b/packages/lu/src/parser/qna/qnamaker/qnaConverter.js
@@ -27,8 +27,8 @@ const qnaToLuContent = function(qnaJSON){
     }
     
     root.forEach(function(qnaItem) {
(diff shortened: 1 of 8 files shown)


vishwacsena:
@vipeketi can you please take a look at this change as well? This bears an impact on the import format to QnA maker in that the ID will now appear after the source. Need update to the download as well as need to ensure replace API supports it as well.

vishwacsena:
@vipeketi can you comment if you can take this change on the QnA service side as well?

vipeketi:
> @vipeketi can you comment if you can take this change on the QnA service side as well?

@vishwacsena  I have made changes for our implementation as below. Hope it works.
// Definition bug fixed for swapping IdMark and SourceInfo Order 
qnaDefinition
    : qnaIdMark? qnaSourceInfo? qnaQuestion moreQuestionsBody qnaAnswerBody promptSection?
    ;

Issue #713, opened 2020-04-16
cross-train: The cross-train config doesn't support multiple intent triggers -> 1 dialog


lei9444:
## Describe the bug
Give a clear and concise description of what the bug is.

Example
Main.dialog has 2 triggers
OnIntent(Name=IntentA) in this trigger, started a dialog called DialogA
OnIntent(Name=IntentB) in this trigger, started a dialog called DialogA

so the config should be like: 
`
        'main.lu': {
          'dialogA.lu': ['IntentA', 'IntentB']
        },
`

## Expected behavior
Give a clear and concise description of what you expected to happen.

## Additional context

[bug]

feich-ms:
close this issue since the fix is merged and verified.

PR #714, opened 2020-04-16
support multi trigger intents point to same lu file

feich-ms:
fix issue #713 from composer. Main changes are supporting multi trigger intents mapping to same lu file in config and logic. Also added tests to cover the scenario. Considering such cases,

triggerIntentA -> a.lu
triggerIntentB -> a.lu
triggerIntentC -> b.lu

After cross train, interruption intent in a.lu will only include utterances under triggerIntentC.


feich-ms:
@boydc2014 fixed: 
1. align the config format between cli and core library.
2. adjust the test cases.
3. adjust cli branch https://github.com/microsoft/botframework-cli/tree/cross-train based on the new changes.

PR #718, opened 2020-04-17
[luis build] change string format in setting and luis recognizer

feich-ms:
Background: SDK side, expression syntax no longer support "-" in the middle of the name, so we change to "_" instead.

change string format in settings and luis recognizer from 'a-1.en-us.lu' to 'a_1.en_us.lu' as string with '-' will be treated as invalid expression string in expression engine


boydc2014:
Looks good to me

PR #730, opened 2020-04-23
Updates to LU round tripping with luis_schema_version 7.0.0

vishwacsena:
- [x] LU -> JSON as well as JSON -> LU now support luis_schema_version 7.0.0
- [x] luis_schema_version 7.0.0 is now the default. Documents (LU or JSON) with 6.0.0 will be upgraded to 7.0.0
- [x] Ability to define prebuilt entities via `@ prebuiltType entityName` notation. Under the cover this will be transformed (unless not within an nDepth entity) to prebuilt type with a role for entityName.
- [x] Updated all tests to 7.0.0 format
- [x] Added new tests to cover upgrade paths from older schema version to 7.0.0


cleemullins:
@munozemilio, I don't think I'm qualified to signoff on this one. Can you take a look?

packages/lu/src/parser/luis/luConverter.js
```diff
@@ -414,7 +421,14 @@ const addRolesAndFeatures = function(entity) {
     let featuresList = new Array();
     entity.features.forEach(item => {
         if (item.featureName) featuresList.push(item.featureName);
-        if (item.modelName) featuresList.push(item.modelName);
+        if (item.modelName) {
+            if (item.isRequired !== undefined) {
```
munozemilio, line 425: Is this the same?

```
            if (item.isRequired === undefined || 
            (item.isRequired !== undefined && 
            item.isRequired !== true)) {
                featuresList.push(item.modelName);
            }
```

packages/lu/src/parser/utils/helpers.js
```diff
@@ -163,36 +164,213 @@ const helpers = {
...
+                if (newList.indexOf(item) === -1) newList.push(item)
+            })
+            entity.children = newList.map(item => JSON.parse(item))
+            deDupeChildren(entity.children)
+        }
+    })
+}
+
+const isRootEntity = function (entityName, entitiesCollection) {
+    if ((entitiesCollection || []).find(ecEntity => ecEntity.name == entityName) !== undefined)
```
munozemilio, line 315: return ((entitiesCollection || []).find(ecEntity => ecEntity.name == entityName) !== undefined) directly?

PR #731, opened 2020-04-24
luis build support to extract authoringKey set by config:set

### LINKED ISSUE #666: luis:build expects authoringKey not subscriptionKey and config doesn't
## Describe the bug
luis:build expects --authoringKey, but config:set:luis only defines --subscriptionKey.  Config should support --authoringKey and luis:build should use it.  

[bug]

feich-ms:
Fix issue #666. After the fix, users can use **_bf config:set -k authoringKey -v xxxxxx_** to set the authoring key in config and luis build will extract the key from config if it is already set. After set users don't need to always provide authoring key in luis build command.


packages/luis/src/commands/luis/build.ts
```diff
@@ -66,6 +68,12 @@ export default class LuisBuild extends Command {
         throw new CLIError('Missing input. Please use stdin or pass a file or folder location with --in flag')
       }
 
+      if (!flags.authoringKey && (!userConfig || !userConfig.authoringKey)) {
+        throw new CLIError('Missing luis authoring key. Please pass authoring key with --authoringKey flag')
```
chrimc62, line 72: >luis [](start = 36, length = 4)

Should be LUIS.

packages/luis/src/commands/luis/build.ts
```diff
@@ -66,6 +68,12 @@ export default class LuisBuild extends Command {
...
+      } else {
+        flags.authoringKey = flags.authoringKey || userConfig.authoringKey
```
chrimc62, line 74: >userConfig.authoringKey [](start = 51, length = 23)

Is there a separate PR for adding this to config?  Even if config allows anything, I would still expect to see documentation updates both for this command and for config.

PR #737, opened 2020-05-01
Update merge for better composer support, error checking, compactness and correctness

### LINKED ISSUE #722: dialog:merge should not expand $ref
Today, dialog:merge expands all $ref references.  This makes the schema file be very large because everywhere something like the adaptive card schema is brought in the definition is duplicated.  In order to prevent that duplication, dialog:merge will replace them with a single in-file definition.  Additionally this will require keeping allof in place.  In order to consume the schema file, you might want to use the npm packages json-schema-ref-parser and json-schema-merge-allof which gives you a simpler schema object.  
[enhancement]

chrimc62:
This is a substantial rewrite of dialog:merge that:
* Dramatically improves error checking
* Ensures the schema is simpler for composer to consume
* Adds support for following .csproj project references
* Ensures included schemas only appear once in schema
* Adds support for schema: protocol to make it easier to refer to common definitions
* Makes merging a library
* Improved testing coverage
Fixes #722


.vscode/launch.json
```diff
@@ -4,24 +4,6 @@
...
-            "name": "Public bf cli",
-            "program": "C:/Users/chrimc/AppData/Roaming/npm/node_modules/@microsoft/botframework-cli/bin/run",
-            "args": [
-                "dialog:generate",
-                "c:/tmp/sandwich/sandwich.schema",
-                "-o",
-                "${env:TEMP}/cli.out"
-            ],
-            "outFiles": [
-                "C:/Users/chrimc/AppData/Roaming/npm/node_modules/@microsoft/botframework-cli/node_modules/@microsoft/bf-dialog/lib"
```
munozemilio, line 19: This test will only work in one machine :)

packages/dialog/src/library/schemaMerger.ts
```diff
@@ -0,0 +1,906 @@
...
+            await fs.remove(this.output + '.expanded')
+
+            this.log(`Finding component .schema files in${os.EOL}  ${this.patterns.join(os.EOL + '  ')}`)
+            for await (const path of this.expandPackages(await glob(this.patterns))) {
+                componentPaths.push(path)
+            }
+
+            if (componentPaths.length === 0) {
+                return false
+            } else {
```
munozemilio, line 122: else not needed

packages/dialog/src/library/schemaMerger.ts
```diff
@@ -0,0 +1,906 @@
...
+                        let end = process.hrtime(start)[1] / 1000000000
+                        if (this.verbose) {
+                            this.log(`Expanding all $ref took ${end} seconds`)
+                        }
+
+                        this.log(`Writing ${this.output}`)
+                        await fs.writeJSON(this.output, finalSchema, this.jsonOptions)
+                    }
+                }
+                if (this.failed) {
```
munozemilio, line 230: you can switch this with the previous if (!this.failed) and return then there is no need to check twice

packages/dialog/src/library/schemaMerger.ts
```diff
@@ -0,0 +1,906 @@
...
+
+    // Convrert local references to absolute so we can keep them as references when combined
+    relativeToAbsoluteRefs(schema: object, path: string) {
+        walkJSON(schema, val => {
+            if (val.$ref) {
+                val.$ref = this.toAbsoluteRef(val.$ref, path)
+            } else if (val.$schema) {
+                val.$schema = this.toAbsoluteRef(val.$schema, path)
+            }
+            return false
```
munozemilio, line 266: why always return false?

packages/dialog/src/library/schemaMerger.ts
```diff
@@ -0,0 +1,906 @@
...
+        } else {
+            path = path.replace(/\//g, ppath.sep)
+        }
+        return ppath.normalize(path)
+    }
+
+    // Expand .csproj packages and projects
+    async expandCSProj(path: string): Promise<string[]> {
+        let references: string[] = []
+        if (!this.projects[path]) {
```
munozemilio, line 302: Reverting condition will reduce the nesting:

if (this.projects[path]) return references

Same can be applied across this function to simplify it

packages/dialog/src/library/schemaMerger.ts
chrimc62: I do need to check twice.  It is a global status that can change in the other block.

---
In reply to: [418591192](https://github.com/microsoft/botframework-cli/pull/737#discussion_r418591192) [](ancestors = 418591192)

packages/dialog/src/library/schemaMerger.ts
chrimc62: It means "keep going".  In other places I return true which means "don't go deeper".

---
In reply to: [418593766](https://github.com/microsoft/botframework-cli/pull/737#discussion_r418593766) [](ancestors = 418593766)

packages/dialog/src/library/schemaMerger.ts
chrimc62: I much prefer single exit functions--easier to understand and debug.  I've been doing more of early exit, but I'm not that fond of that style.

---
In reply to: [418604887](https://github.com/microsoft/botframework-cli/pull/737#discussion_r418604887) [](ancestors = 418604887)

PR #759, opened 2020-05-06
fix pattern issue in cross train core lib

feich-ms:
Fix issue reported in composer here microsoft/BotFramework-Composer#2749. It was marked as P0 of R9 RC just recently(5.5). If we think this is release blocker for composer, we should make it in.
The main changes are removing all patterns with prebuilt entity as patternany entity in Interruption intent. For other patterns, we still keep them in Interruption intent.

### DIFF
--- a/packages/lu/src/parser/cross-train/crossTrainer.js
+++ b/packages/lu/src/parser/cross-train/crossTrainer.js
@@ -12,6 +12,7 @@ const DiagnosticSeverity = require('../lufile/diagnostic').DiagnosticSeverity
 const fileHelper = require('../../utils/filehelper')
 const exception = require('../utils/exception')
 const retCode = require('../utils/enums/CLI-errors')
+const prebuiltEntityTypes = require('../utils/enums/luisbuiltintypes').consolidatedList
 const NEWLINE = require('os').EOL
 const path = require('path')
 const QNA_GENERIC_SOURCE = "custom editorial"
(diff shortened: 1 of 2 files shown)


feich-ms:
Please help to take this PR to next RC as it is resolving an issue tagged as P0 by composer today @munozemilio @vishwacsena.

packages/lu/src/parser/cross-train/crossTrainer.js
```diff
@@ -300,21 +301,20 @@ const removeDupUtterances = function (resource) {
...
 const extractIntentUtterances = function(resource, intentName) {
   const intentSections = resource.Sections.filter(s => s.SectionType === LUSectionTypes.SIMPLEINTENTSECTION || s.SectionType === LUSectionTypes.NESTEDINTENTSECTION)
-  const curlyRe = /.*\{.*\}.*/
 
   let intentUtterances = []
   if (intentName && intentName !== '') {
     const specificSections = intentSections.filter(s => s.Name === intentName)
     if (specificSections.length > 0) {
-      intentUtterances = intentUtterances.concat(specificSections[0].UtteranceAndEntitiesMap.map(u => u.utterance).filter(i => curlyRe.exec(i) === null))
+      intentUtterances = intentUtterances.concat(specificSections[0].UtteranceAndEntitiesMap.map(u => u.utterance))
```
beyackle, line 309: Is there a reason we can't just change these (here and on lines 314 and 317) to `intentUtterances.push()` instead of making a whole new array and reassigning the variable to it?

packages/lu/src/parser/cross-train/crossTrainer.js
```diff
@@ -440,8 +440,11 @@ const qnaCrossTrainCore = function (luResource, qnaResource, fileName, interrupt
     trainedQnaResource = new SectionOperator(new LUResource([], modelInforContent, [])).addSection(qnaContents)
   }
 
+  // remove utterances with curly brackets
+  const utterancesWithoutPatterns = utterances.filter(i => /{([^}]+)}/g.exec(i) === null)
```
beyackle, line 444: This filter could be written as `i => ! ( /{([^}]+)}/g.test(i) )`.

PR #763, opened 2020-05-07
fix lu validation bugs for empty comments and single curly bracket

feich-ms:
Fix minor lu validation issues reported by composer https://github.com/microsoft/BotFramework-Composer/issues/2910 and https://github.com/microsoft/BotFramework-Composer/issues/2940. The fix for 2910 is simple(first commit), just loose the rule to allow comment be empty. The fix for 2940 is also simple(second commit), just allow single curly brackets in utternace. rush test of lu and qna all passed. 2910 is marked R9-RC in composer side, so I think this PR should go into BF-CLI next RC.


boydc2014:
LGTM, cc @munozemilio @vishwacsena

PR #781, opened 2020-05-12
fix section operator bug

feich-ms:
Fix section operator bug reported by @chrimc62. This issue was introduced by another PR for fixing a composer issue. After fix, the operations for other sections out of intent sections will work.


boydc2014:
We should target this into master.

feich-ms:
Will create a new one targeting master. Close this first.

PR #784, opened 2020-05-13
Update dialog:merge tests around projects

chrimc62:
Also fix a bug where npm semver rather than nuget semver was used.


.vscode/launch.json
```diff
@@ -200,6 +201,26 @@
...
+            ],
+            "args": [
+                "dialog:merge",
+                "runbot.csproj",
+                "-o",
+                "runbot.schema",
+                "--verbose"
+            ],
+            "internalConsoleOptions": "openOnSessionStart",
+            "cwd": "${workspaceFolder}/../botbuilder-samples/experimental/generation/runbot/"
```
munozemilio, line 222: This path seems to be very specific

Issue #770, opened 2020-05-08
luis:build parse lu file with incorrect locale priority


feich-ms:
## Versions
What CLI version are you using.
What Nodejs version are you using
What command-line interpreters are you using
What OS are you using

## Describe the bug
When users use luis:build to publish lu files, the locale of that lu file will be specified under such priority
Culture from file name(a.en-us.lu) > Culture from lu file setting(@> !# @app.culture = en-us) > default culture(default to en-us if not specified by users). 

However, in current implementation, the culture passed to build function in luisCollate is default culture instead of culture from file name. This will introduce parser errors when default culture is set to other unreasonable values.

## To Reproduce
1. use luis:build to publish a lu file with below definitions but set defaultCulture to it-it
```@ prebuilt personName```
2. You will see some warings and errors like below
<pre>
[WARN]: Requested PREBUILT entity "personName is not available for the requested locale: it-it
  Skipping this prebuilt entity..
</pre>

<pre>
The model name { personName } are reserved.
</pre>

## Expected behavior
Give a clear and concise description of what you expected to happen.

## Screenshots
![image](https://user-images.githubusercontent.com/43032123/81394081-fcee2500-9153-11ea-8c7c-66f154ee105f.png)

## Additional context

[bug]

Issue #778, opened 2020-05-11
Support chinaeast2 region in luis build


feich-ms:
**Is your feature request related to a problem? Please describe.**
In luis build cmd, chinaeast2 region is targetting `https://chinaeast2.api.cognitive.azure.cn` instead of `https://${region}.api.cognitive.microsoft.com`. To support this, luis builder interface need to be changed.

**Describe the solution you'd like**
Change interface of luis builder function.

**Describe alternatives you've considered**
Change interface of luis builder function.

**Additional context**

[enhancement]

PR #785, opened 2020-05-13
[R10] fix issues #770 and #778

feich-ms:
fix issue #770 and issue #778.


vishwacsena:
@feich-ms can we add tests?

feich-ms:
@vishwacsena, unit tests for locale issue are added but the endpoint feature cannot be tested by unit tests as it needs to be with config:set:luis command. I did manual tests in package/cli/bin folder with below cmd
1. run config:set:luis --endpoint https://chinaeast2.api.cognitive.azure.cn
2. run luis:build --in a.zh-cn.lu --authoringKey xxx --out . --suffix feich --botName test --region chinaeast2 --fallbackLocale zh-cn

And it works well when I'm using the authoringKey from https://luis.azure.cn/

PR #793, opened 2020-05-17
Remove unnecessary validation that required child entity names to be globally unique

### LINKED ISSUE #791: Entities need to be unique within their containing scope
This should not fail

```
@ prebuilt number hasRoles Quantity
 
@ ml foo = 
 - @ number number
 ```

### PR DESCRIPTION
**Pull request by vishwacsena**
- [x] Fixes #791 
- [x] Removed unnecessary tests.

### CODE REVIEW

File: packages/lu/test/parser/lufile/parseFileContents.nDepthEntity.test.js [hunk 1]
```diff
@@ -204,18 +204,6 @@ describe('V2 NDepth definitions using @ notation', function () {
...
-        let luFile = `
-            @ml xyz
-            @ml xyz1 = 
-                - @ ml xyz
-
-        `;
-        parseFile.parseFile(luFile)
-            .then(res => done(res))
-            .catch(err => done())
-    });
```
chrimc62, line 217: Wouldn't it be better to keep the test and switch expectations?

Issue #790, opened 2020-05-17
Need a way to get reason of error "No LU or Luis content parsed!"


dawwa:
## Versions
@microsoft/botframework-cli/4.9.0 win32-x64 node-v12.16.3
Windows Server 2019

## Describe the bug
We see bf luis:convert fails when running in AzureDevOps Build Pipeline, but does not repro when running the same command locally.

I did a brief check on the source code, looks like this, is there some workaround we can get the reason why the result is empty?

```ts
        result = await LuisBuilder.build(luFiles, flags.log, flags.culture)
        if (!hasContent(result)) {
          throw new CLIError('No LU or Luis content parsed!')
        }
```

## Current behavior
We only see "No LU or Luis content parsed!" as one line error in the build log, but have no idea why it happens as we could not repro it locally.

## Expected behavior
We need to understand what's the real failure behind the error message.

vishwacsena:
@dawwa we will work on improving the error messages. To get yourself unblocked, can you check to see if the files exist and are accessible to the CLI in your pipeline? This usually occurs when we read an empty file or a file that does not contain any LU content.

dawwa:
@vishwacsena , @feich-ms helped to produce a private bit which prints more details, that helped discover the issue.

**TL;DR:**
MSBuild Exec interfered the stdin parameter of bf CLI with unexpected token  `∩╗┐` 
The proposed fix is to ignore stdin when --in option is used.
@feich-ms will push a fix into nightly feed until an official fix rolls out.

**Detail Rootcause**
Basically, the stdin is interfered by MSBUILD Target Exec command, which is with System.Text.UTF8 encoding, and by default, UTF-8 encodings in .NET use a Unicode Byte Order Mark at the beginning of the stream, which caused the `∩╗┐` we see from the stdin that bf CLI received.

Similar issues of other products mentioned about this problem here
https://github.com/microsoft/msbuild/issues/397
https://github.com/TypeStrong/typedoc/issues/133

feich-ms:
Created a PR here https://github.com/microsoft/botframework-cli/pull/796

feich-ms:
Verfied in my private feed that this issue was resolved https://msazure.visualstudio.com/One/_build/results?buildId=31208400&view=logs&j=fd490c07-0b22-5182-fac9-6d67fe1e939b&t=dcf93f35-c1ba-5dce-0078-8d628250da3a&l=1005

feich-ms:
fixed by PR #796

[...]

PR #796, opened 2020-05-18
fix the priority of stdin and in

feich-ms:
fix issue #790


packages/lu/src/utils/filehelper.ts
```diff
@@ -17,15 +17,15 @@ const globby = require('globby')
 
 export async function getLuObjects(stdin: string, input: string | undefined, recurse = false, extType: string | undefined) {
   let luObjects: any = []
-  if (stdin) {
-    luObjects.push(new luObject(stdin, new LUOptions('stdin')))
-  } else {
+  if (input) {
```
munozemilio, line 20: If I remember correctly stdin takes precedence over input in other commands. We should have the same behavior across the commands...@vishwacsena thoughts?

feich-ms: How about making input precedent over stdin in all cli commands to keep the consistency?  It seems blocking issue for customers if we don't take this change. Do you have better solutions? @munozemilio @vishwacsena

vishwacsena: Looking at the issue we are attempting to fix, I agree we should have `--in` supersede anything passed in via `stdin`. @feich-ms, can we also via this PR update chatdown:convert as well as qnamaker:convert to prefer `--in` over `stdin`? Also do this for `lg` commands? I cannot think of other places where we support reading from stdin but would be great if you can scan the BF-CLI code and update all places.

feich-ms: sure. will make changes for all occurrences.

feich-ms: All occurrences are resolved

vishwacsena:
@feich-ms this seems to be missing updates to chatdown..?

feich-ms:
@vishwacsena, I scanned the chatdown code, found that the --in already has higher priority than stdin there and lg tool does not accept stdin.

feich-ms:
@vishwacsena, does this look good to go for you?

PR #803, opened 2020-05-19
Update BF-LU to update validation for entity references in patterns

vishwacsena:
Fixing an issue that @chrimc62 found via generated dialogs.

### DIFF
--- a/packages/lu/src/parser/utils/helpers.js
+++ b/packages/lu/src/parser/utils/helpers.js
@@ -215,6 +215,8 @@ const updateToV7 = function(finalLUISJSON) {
 const verifyPatternsDoNotHaveChildEntityReferences = function(finalLUISJSON, entityParentTree)
 {
     if (finalLUISJSON.patterns === undefined || !Array.isArray(finalLUISJSON.patterns) || finalLUISJSON.patterns.length === 0) return;
+    // update entityParentTree with all entity types.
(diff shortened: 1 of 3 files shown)


packages/lu/src/parser/utils/helpers.js
```diff
@@ -234,6 +238,26 @@ const verifyPatternsDoNotHaveChildEntityReferences = function(finalLUISJSON, ent
     })
 }
 
+const updateEntityParentTreeWithAllEntityTypes = function(finalLUISJSON, entityParentTree)
+{
+    (finalLUISJSON.prebuiltEntities || []).forEach(entity => addEntityToParentTree(entityParentTree, entity.name));
```
munozemilio, line 243: The same operation is done for all properties. This could be replaced by a for going over an array

munozemilio:
@vishwac could you please add details on the fix priority and issue

vishwacsena:
@munozemilio sure. LUIS recently relaxed the requirement around uniqueness of entity names. Prior to build, all entity names across their scope (including child entities) need to be unique across the entire application. 

With recent changes to LUIS, entity names need to be unique under the immediate parent scope. This means I can have a global entity `foo` and a child entity with the same name. 

Without this fix, BF-LU parser will disallow child entities with same name as a different globally scoped entity. e.g. 

```
@ list foo
@ ml parent = 
    - @ foo foo

# test
- this is a pattern with {foo}
```

PR #804, opened 2020-05-20
resolve luis build issues

### LINKED ISSUE #800: Issues of build and cross-train cli
## Versions
What OS are you using

## Describe the bug
Luis:build
1.	Luis:build and qnamaker:build do not write .dialog if no –out is specified.
2.	–suffix when not specified resulted in LUIS app with (Undefined). 
3.	Verify luis:build detects changes to references.
4.	Fix issue with config relative path. Relative path must always be relative to the CWD()5
5.	C:\users\temp> bf luis:build –in . –out generated –config ..\foo.config.json
 
Luis:cross-train
1.	-- config is relative to –in and not relative to the pwd()
2.	Cross-train config is file name case sensitive. 
3.	Cross-train should only copy over files specified in the cross-train config to output but all references should be fully resolved before they are copied over.
4.	QnA meta-data property is not applied to references
5.	Apply cross-train for .lu and .qna documents after fully resolving imports. 
6.	If multiple source .lu or .qna files are pulling in the same reference, the for QnA, meta-data pairs need to be added to the same QnA pair (e.g. root as well as dialog A pulling in some chitchat utterances) 
 
Qnamaker:build
1.	There is no way to get endpoint key. We should print this on screen. Note the endpoint key is different from authoring key and I believe is per KB – need to verify and resolve.
2.	Fix issue with config relative path. Relative path must always be relative to the CWD()

## Expected behavior

## Additional context

[bug]

feich-ms:
Resolve luis build related issues in #800 

- Luis:build

   - [x] Luis:build do not write .dialog if no –out is specified.
   **_Resolved_**: fixed in PR
   - [x] suffix when not specified resulted in LUIS app with (Undefined).
   **_Resolved_**: fixed in PR
   - [x] Verify luis:build detects changes to references.
   **_Resolved: verified that luis:build did detect changes of references since it's merging all references into one and do comparison. No fix is needed._**
   - [x] Fix issue with --config relative path. Relative path must always be relative to the CWD()
   _**Resolved: It is --luConfig in luis build and it is already related to cwd. No fix is needed.**_


boydc2014:
Maybe break into 3 prs resolving luis:build, qna:build, crosstrain?

feich-ms:
sure. This one will resolve luis build issues.

Issue #751, opened 2020-05-05
bf qnamaker:build should support --qnaconfig


chrimc62:
**Is your feature request related to a problem? Please describe.**
luis:build --luconfig allows you to capture parameters in a generated file and gives more control.  qnamaker supports the same parameters and has the same needs, but does not support --qnaconfig equivalent.  

**Describe the solution you'd like**
Add --qnaconfig to qnamaker:build.

[enhancement]

PR #812, opened 2020-05-20
resolve qnamaker build issues

feich-ms:
Resolve qnamaker build related issues in #800
- Qnamaker:build
    - [x] Qnamaker:build do not write .dialog if no –out is specified.
   **_Resolved_**: fixed in PR

   - [x] There is no way to get endpoint key. We should print this on screen. Note the endpoint key is different from authoring key and I believe is per KB – need to verify and resolve.
**_Resolved: fixed in this PR_**

   - [x] Fix issue with config relative path. Relative path must always be relative to the CWD()
_**Resolved: --qnaConfig is not supported yet. We have a feature request here #751. Will make it relative to CWD() in that implementation.**_


vishwacsena:
@feich-ms can we also update this so we do not write out `dialog` files unless `--out` is specified?

feich-ms:
Actually I already did this in pr #804 when updating Luis build. will move from that pr to this one to help better review

PR #825, opened 2020-05-27
fix lu parser error when invalid square brackets occur in utterances

### LINKED ISSUE #795: lu parser issue editing Luis intent mess up the file
## Versions
What OS are you using

## Describe the bug
reported by composer https://github.com/microsoft/BotFramework-Composer/issues/3071

## Expected behavior

## Additional context

[bug]

feich-ms:
fix #795. Before the fix, the square brackets occur after an intent will be treated as import section which caused the intent body missed the error intent line. After the fix, the invalid square brackets will be treated as error intent line instead of import section so that it can be included in the body of the intent.


packages/lu/test/parser/lufile/sectionapi.test.js
```diff
@@ -269,4 +269,22 @@ describe('Section CRUD tests for entity', () => {
...
+
+describe('Section CRUD tests for error import in utterances', () => {
+    let luresource = undefined;
+    
+    it('init section test', () => {
+        let fileContent = 
+`# Greeting
+- hi
+[]
+> @ ml city`;
```
zhixzhan, line 282: suggest add, ensure, `#hello` not included in `# Greeting`
`# Greeting
 - hi
 []
 > @ ml city
# hello
- halo
`

feich-ms: fixed. Add more corner cases to cover above scenario

feich-ms:
@vishwacsema @munozemilio could you please help to take a look at this PR?

vishwacsena:
Not sure I follow why this is not an error .. can you clarify? 
```markdown
[]
[
]
```

feich-ms:
@vishwacsena. It is 100% error and it will throw out. The changes here is to make sure the next valid intent section is not impacted by this error when parsing. Composer is using the lu parser in their lu editor. They want lu parser to have the capability of parsing with errors which means error text in current section can be pased into Body (even though the errors are caught)and errors in current section doesn't impact next valid section parsing. For example,

Considering below lu content

```
`# Greeting
- hi
[]
# Cancel
- cancel that`;
```

The parsed lu resource will have error of `Invalid intent body line, did you miss \'-\' at line begin` but the Body of Greeting intent will be `- hi$\r\n[]` and Cancel intent can be parsed correctly.

PR #830, opened 2020-05-28
add retry for luis rate limiting issue

### LINKED ISSUE #819: Add retry mechanism if luis build run into rate limiting issue
## Versions
What OS are you using

## Describe the bug
image

## Expected behavior

## Additional context

[bug]

feich-ms:
fixes #819. 
For bf-lu, exposed multiple parameters for composer or other client to consume to configure the luis api call rate:
1. **Luis api tps(concurrent number)**: default to 5 and can switch to 50 when using a paid key
2. **time bucket of concurrent api calls**: default to 1100 ms and can switch to higher value to reduce rate limit exception frequency
3. **retry count:** default to 1 and can switch to higher value to reduce rate limit exception frequency
4. **retry delay time:** default to 1000ms and can switch to higher value to redeuce rate limit exception frequency

For bf-luis-cli, all values of above parameters are default ones which means adding once retry after 1 second for any rate limiting error when calling luis api. If still running into this issue after retry, will let it throw out. Tested it manually from luis:build command, and it can mitigate the error rate limiting frequency effectively but cannot guarantee 100% resolving.

Composer can configure the paramters based on user feedbacks.


boydc2014:
can we make the parameters in the scheduler configrable like
* retry count
* delay time
* time bucket per request

feich-ms:
resolved Dong's suggestions.

PR #840, opened 2020-06-04
qnamaker:build support --qnaConfig

### LINKED ISSUE #751: bf qnamaker:build should support --qnaconfig

feich-ms:
fix #751 to support --qnaConfig to configure the cli paramters. It also support to auto-extract subscriptionKey set by bf config:set:qnamaker.


packages/qnamaker/src/commands/qnamaker/build.ts
```diff
@@ -27,59 +30,89 @@ export default class QnamakerBuild extends Command {
...
       flags.stdin = await this.readStdin()
 
-      if (!flags.stdin && !flags.in) {
+      if (!flags.stdin && !inVal && files.length === 0) {
         throw new CLIError('Missing input. Please use stdin or pass a file or folder location with --in flag')
       }
 
-      if (flags.dialog && flags.dialog !== recognizerType.MULTILANGUAGE && flags.dialog !== recognizerType.CROSSTRAINED) {
+      if (!subscriptionKey) {
+        throw new CLIError('Missing QnAMaker subscription key. Please pass subscription key with --subscriptionKey flag or specify via bf config:set:qnamaker.')
```
vishwacsena, line 82: bf config:set:qnamaker --subscriptionKey <your subscription key>

packages/qnamaker/src/commands/qnamaker/build.ts
```diff
@@ -27,59 +30,89 @@ export default class QnamakerBuild extends Command {
...
+      }
+
+      if (!botName) {
+        throw new CLIError('Missing bot name. Please pass bot name with --botName flag.')
```
vishwacsena, line 86: Or specify via --qnaConfig

feich-ms: resolved

packages/qnamaker/src/commands/qnamaker/build.ts
feich-ms: resolved

PR #845, opened 2020-06-08
support endpoint via arg or cli config for luis and qnamaker build

### LINKED ISSUE #843: luis:build should allow endpoint via arg or use from cli config
luis:build today does not allow endpoint configuration via CLI arg. This is necessary when user is creating a authoring subscription via azure portal. Of course this key works with the region based mapping but I ran into an issue when I deleted a prior LUIS resource and I could not get luis:build to work because it kept using an invalid and old endpoint I had set via bf config:set:luis. 

We should allow explicit override of `endpoint` (authoring endpoint) to the `luis:build` command and if none specified, we should fall back to any specified via `bf config:set:luis`.

feich-ms:
fix #843

### DIFF
--- a/packages/luis/README.md
+++ b/packages/luis/README.md
@@ -368,6 +368,8 @@ OPTIONS
   --suffix=suffix                  Environment name as a suffix identifier to include in LUIS app name. Defaults to
                                    current logged in user alias
 
+  --endpoint                       Luis authoring endpoint for publishing
+
 EXAMPLE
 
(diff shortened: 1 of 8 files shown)


vishwacsena:
Does this mean they are also configurable via luconfig?
Also can we change --luconfig to --qnaconfig for qnamaker:build?

feich-ms:
@vishwacsena, yes, they are configurable via luConfig or qnaConfig. Just added the test cases to cover that. The typo of luConfig in qnamaker:build README is also resolved.

PR #852, opened 2020-06-11
Support merging .uischema files

chrimc62:
This adds to dialog:merge the ability to merge .uischema files to implement #852 .
To do this it walks-over the component (.csproj, package.json, .nuspec, nuget, packages.config) tree and builds a breadth-first parent before child order to use for when merging .uischema definitions.  The uischema files are checked to ensure they are valid according to the uischema definition and to ensure that all properties are found in the schema.  The result is a merged ui schema file per locale like <schemaName>[.locale].uischema.

It will also do an analysis and show errors in content files of:
1) Multiple of the same  asset like (.lu, .lg, .qna, .schema, .uischema, .dialog) within a component.
2) Different components that define the same .schema file.

Also added the ability to infer the output file name from the first project.

NOTE: I did keep the ability to do .schema files directly as patterns.  This is mainly for testing.


packages/dialog/README.md
```diff
@@ -13,9 +13,12 @@ This package is intended for Microsoft use only and should be consumed through @
 
 # Commands
 <!-- commands -->
-* [`bf dialog`](#bf-dialog)
-* [`bf dialog:merge PATTERNS`](#bf-dialogmerge-patterns)
-* [`bf dialog:verify PATTERNS`](#bf-dialogverify-patterns)
+- [@microsoft/bf-dialog](#microsoftbf-dialog)
```
munozemilio, line 16: This should be outside the commands toc

chrimc62: I did auto-gen and then restored things from the original.  For example the slashes for code references are still the wrong way.

PR #854, opened 2020-06-12
CRUD optimizations for qna sections

feich-ms:
Optimize CRUD operations for qna sections which will be consumed by composer qna editor.


liweitian:
Found some problems on it. Need to discuss with @feich-ms . Do not merge it at this time.

PR #856, opened 2020-06-12
Add docs for cross-train command for both luis and qnamaker

vishwacsena:


feich-ms:
Looks good to me. The only comment is, for the cross-train.png in qnamaker docs, the _interruption intent of cross-trained child1.lu and child2.lu should also include the questions from rootDialog.qna.

PR #858, opened 2020-06-15
add id for generated recognizer and multiLang recognizer

### LINKED ISSUE #853: For Cross-Train, require id property in both lu.dialog and qna.dialog files
When doing cross-train, it checks each recognizer id by the function EnsureRecognizerIds() in CrossTrainedRecognizerSet.cs. So we need to add the id property in each lu.dialog or qna.dialog file, including lu.dialog, en-us.lu.dialog, qna.dialog and en-us.qna.dialog.

feich-ms:
fix #853 to add id for generated recognizer and multiLang recognizer dialog files which are required in SDK when including luis and qnamaker cross-train.


feich-ms:
@vishwacsena, sure, I will work with zongyang @coldplaying42 to do manual tests in declarative bot.

feich-ms:
@vishwacsena, I have tried to generate the dialog assets with id property and compare them with the hand-crafted dialog files @coldplaying42 was using to test lu and qna cross-train runtime. They are exactly same and can work with the runtime.

chrimc62:
I'm confused why this is necessary.  By design the id of a top-level instance is the filename which is required to be globally unique.  In fact, I believe dialog:verify will complain if a top-level id is present since that sets up a conflict between the filename and the id.  The reason id exists is to name a $kind that is inside another one.

feich-ms:
Hi @chrimc62, the id added by this PR is for CrossTrainedRecognizer to find the right recognizer when `DefersToRecognizer_{Id}` is the top intent for a user input. I went through the code here https://github.com/microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Builder.Dialogs.Adaptive/Recognizers/CrossTrainedRecognizerSet.cs#L99. It is using the `LUIS_{fileName}` and `QnA_{fileName}` to map the luis/qna recognizers into a dictionary. When `DefersToRecognizer_{Id}` is the top intent, it will get the `{id}` value and find the recognizer in the dictionary. If we only have luis recognizer, we can have filename as the id, but if we have both luis and qna recognizer, filename may not be globally unique.

If we don't use `LUIS_{fileName}` or `QnA_{fileName}` as recognizer id, we may need to make changes in both above runtime logic and luis/qnamaker:build cli. For example, we don't write out id in luis/qnamaker:build recognizers but instead we can add `LUIS_` or 'QnA' prefix in runtime when loading recognizer files of luis and qna since we have $kind property and filename(a.lu.dialog vs a.qna.dialog) which can tell us this is luis or qna recognizers. @chrimc62 @vishwacsena, do you have any suggestions?

chrimc62:
OK, we removed the old id so this is the right one.

---
In reply to: [645227009](https://github.com/microsoft/botframework-cli/pull/858#issuecomment-645227009) [](ancestors = 645227009)

PR #866, opened 2020-06-23
add .dialog extension for cross-trained qna filter values

feich-ms:
add .dialog extension for cross-trained qna filter values which are required by cross-train runtime

### DIFF
--- a/packages/lu/src/parser/cross-train/crossTrainer.js
+++ b/packages/lu/src/parser/cross-train/crossTrainer.js
@@ -433,7 +433,7 @@ const qnaCrossTrainCore = function (luResource, qnaResource, fileName, interrupt
   // update qna filters
   let qnaSectionContents = []
   for (const qnaSection of qnaSections) {
-    qnaSection.FilterPairs.push({ key: 'dialogName', value: fileName })
+    qnaSection.FilterPairs.push({ key: 'dialogName', value: `${fileName}.dialog` })
     let qnaSectionContent = ''
     if (qnaSection.source !== QNA_GENERIC_SOURCE) {
       qnaSectionContent += `> !# @qna.pair.source = ${qnaSection.source}${NEWLINE}${NEWLINE}`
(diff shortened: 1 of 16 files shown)


vishwacsena:
@feich-ms this would not work with code based bot since there is no `.dialog`. Can we investigate what causes the need to include `.dialog` when the QnA recognizer sends up the meta data filter? This should just be the dialog name without `.dialog`.

feich-ms:
@vishwacsena, I synced with @coldplaying42 who was ever running QnAMakerRecognizer in form bot. He pointed that the QnAMakerRecognizer is adding QnA filter values with `.dialog` extension. The code piece is here https://github.com/microsoft/botbuilder-dotnet/blob/master/libraries/Microsoft.Bot.Builder.AI.QnA/QnAMakerRecognizer.cs#L170.
```
List<Metadata> filters = new List<Metadata>();
if (IncludeDialogNameInMetadata.GetValue(dialogContext.State))
{
     filters.Add(new Metadata() { Name = "dialogName", Value = dialogContext.ActiveDialog.Id });
}
```

Here `dialogContext.ActiveDialog.Id` has `.dialog` extension.
Add @luhan2017 to confirm.

luhan2017:
Yes, I've debugged the code, the "dialogContext.ActiveDialog.Id" is the full id with .dialog extension. As @vishwacsena said, there might be an inconsistency between the code-based adaptive dialog and declarative in the sdk side, let's hold on this PR, I'll debug it on the sdk side, and try to fix it there.

luhan2017:
I've debugged it. The resource id is the full file name with extension and the resource id is assigned to dialog id which also includes .dialog extension. Should we remove the .dialog extension in the assignment?  @vishwacsena 

image

image

vishwacsena:
@luhan2017 to keep this least impactful, let's just fox the QnA recognizer to drop `.dialog` before posting it as meta data filter. 

https://github.com/microsoft/botbuilder-dotnet/blob/59f8ccca5928a3352f0ec815881d59f474bf478b/libraries/Microsoft.Bot.Builder.AI.QnA/QnAMakerRecognizer.cs#L161

luhan2017:
@vishwacsena , let's close this, and I will submit another PR in the sdk side.

PR #870, opened 2020-06-28
[Do Not Merge]luis and qnamaker build support to write out empty crosstrained recognizer

feich-ms:
fix composer issue https://github.com/microsoft/botbuilder-dotnet/pull/4144


feich-ms:
close this one since it is merged into PR https://github.com/microsoft/botframework-cli/pull/889 which depends on this one.

Issue #880, opened 2020-07-08
Move CrossTrainRecognizer generation logic from lubuild/qnabuild to crosstrain 


luhan2017:
we hit some issue when enable qna and crosstrain recognizer in composer. So I propose to have a more clear design of the functionality of each lib. We should not put the crosstrain recognizer logic in qnabuild or lubuild, what if non-composer users and they don't care about crosstrain? I think crosstrain recognizer genration logic should be in crosstrain library. 

The whole crosstrain + lubuild + qnabuild process should look like this.
1. composer will call crosstrain api.
Here is the crosstrain logic.
if .lu and .qna are both empty. do nothing, and just generate a CrossTrainedRecognizer with no recognizers inside.
if only .lu not empty. crosstrain on lu only, and generate a crosstrainedrecognizer with lu recognizer only. 
if only .qna not empty. no need crosstrain, and generate a crosstrainedrecognzier with qna recognizer only.
if both not empty, crosstrain them, and generate a crosstrained recognizer with both lu and qna.

lubuild don't care qna, qnabuild don't care lu.
they just take the .lu files or .qna files if files are not empty, prompt for keys and generate the multilanuage luis/qna recognizer. 
 
@vishwacsena, @boydc2014 thoughts?

feich-ms:
I don't think it's necessary to move CrossTrainRecognizer generation logic to crosstrain library. It will confuse users if we have CrosstTrainedRecognizer and other recogizers generated in different library. I prefer to optimize the lu/qnamaker build apis to provide a function for users to generate empty crosstrained recognizer when both .lu and .qna are empty files. Currently I have three functions exposed in builder.ts. The flow to use them are: 

loadContents() => load lu files and generate empty recognizers including single recognizer, multi language recognizer and crosstrained recognizer

build() => load lu contens and all empty recognizers from loadContents() and fill the recognizers with appid/kbid and other information after luis/qnamaker published, and then generate all contents that need to write out ot print

writeDialogAssets() => write recognizers and settings contents generated from build() function

I think for `.lu and .qna are both empty` scenario, we can make changes in loadContents() to return a flag to indicate whether all lu and qna files are empty, if so, directly call writeDialogAssets() to write empty crosstrained recognizer only.

feich-ms:
Fixed in PR #889

Issue #879, opened 2020-07-08
qnabuild write dialog assets for each dialog instead of writing a shared set


feich-ms:
## Versions
What CLI version are you using.
What Nodejs version are you using
What command-line interpreters are you using
What OS are you using

## Describe the bug
In crosstrain CLI, when cross training qna with luis, the `DeferToRecognizer_QnA_{dialogName}` should change to `DeferToRecognizer_QnA_{botName}` as runtime is trying to find the recognizer by the suffix. Since we are generating dialog with botName as the file name prefix instead of dialogName in qnamaker:build, so for composer, it requires `DeferToRecognizer_QnA_{botName}`. To resolve this, we need to add one more `botName` parameter for luis/qnamaker: crosstrain cmd. What do you think? @vishwacsena 

## Expected behavior
Give a clear and concise description of what you expected to happen.

## Additional context

[bug]

boydc2014:
triaged, this is not the direction, we should follow up with dialog name

feich-ms:
yes, we should not touch dialog name in filter. Instead I made changes in qnamaker build to write out dialog files for each dialog but the recognier pointing to merged KBs based on locales. PR is here https://github.com/microsoft/botframework-cli/pull/889

Issue #878, opened 2020-07-08
bf crosstrain, if qna file is empty, we should not crosstrained for qna and generated an non-empty qna file


luhan2017:
## Versions
What CLI version are you using.
What Nodejs version are you using
What command-line interpreters are you using
What OS are you using

## Describe the bug
@vishwacsena , in composer, the expected behavior is if qna file is empty, user don't need to input a qna authoring key, and the qnabuild should not be called. but we found that the crosstrain did generate a non-empty qna file.

The expected logic should be:
only of lu and qna file both exist, it trains cross the qna and lu.
if lu is empty, just keep qna, and don't do the crosstrain.
if qna is empty, just crosstrain luis, don't write anything into qna file.

## Expected behavior
Give a clear and concise description of what you expected to happen.

## Additional context

[bug]

feich-ms:
Fixed in PR #889

PR #889, opened 2020-07-13
cross train and lu/qna build optimization for composer

### LINKED ISSUE #883: crosstrain should add filter for qna only scenarios
In composer, there is a scenario that only qna is enabled in bot, in such scenario, crosstrain will not be called and dialog filters are not added to qna files. Without

**Describe alternatives you've considered**

**Additional context**

[enhancement]

feich-ms:
fix #883, #880, #879, #878 and https://github.com/microsoft/botbuilder-dotnet/pull/4144. All these issues are related with composer crosstrain and lu/qna:build scenarios. This PR is to resolve these issues including:
1. add filter meta data for all qna pairs
2. adjust the recognizer generation logic for qna to generate recognizer dialog files for each dialog instead of sharing single one of botName, but in settings file, the kb ids of qna recognizers will be same for same locale.
3. Optimize lu/qna builder interface and flow to write out the empty crosstrained recognizer for both lu and qna files are empty scenario
4. Do not add any crosstrained contents to lu or qna files if all lu files or all qna files are empty in crosstrained flow
5. adjust and add Unit tests

Now composer only need to call below flow directly and DO NOT need any other tricy calling logic:
call crosstrain() -> lubuild.loadContents() ->lubuild.build() -> lubuild.writeDialogAssets() -> qnabuild.loadContents() -> qnabuild.build() -> qnabuild.writeDialogAssets()


boydc2014:
@alanlong9278 @luhan2017  can you help verify this feature is as what we expected at composer side

PR #898, opened 2020-07-15
enable to configure luis publishing slot enviroment

### LINKED ISSUE #855: [MCS-DE-GO] -- allow configuring the LUIS slot on lubuild.Builder publish application
When using the luBuilder it would be important to be able to set the slot where the LUIS model is published. Currently it is hardcoded to the production slot. 
Source file: lubuild/core.js  method: publishApplication

 `this.client.apps.publish(appId, {
            versionId,
            isStaging: false
        });`

In our MCS project with Gothaer the customer wants to share a LUIS instance for the DEV and TEST environments by setting the staging slot for DEV and production slot for TEST, hence it is paramount that we are able to publish the .lu files for both slots.

As a workaround we can make a copy of the repo and implement that change in the .ts files, however that is far from an ideal solution as we would lose the ability to easily retrieve the newest updates.

This parameter should be set in the lubuild.Builder build method as this is the one being used by the BF Composer process

@benbrown @goergenj @sgellock

feich-ms:
fix #855 to enable users to configure the publishing slot enviroment. Before this change, lu build will default publish to production slot with no configuration capability. After the fix, users can specify --isStaging in lu:build to publish to staging slot. Default to production slot if not specified.

### DIFF
--- a/packages/lu/src/parser/lubuild/builder.ts
+++ b/packages/lu/src/parser/lubuild/builder.ts
@@ -139,6 +139,7 @@ export class Builder {
     suffix: string,
     fallbackLocale: string,
     deleteOldVersion: boolean,
+    isStaging: boolean,
     multiRecognizers?: Map<string, MultiLanguageRecognizer>,
     settings?: Map<string, Settings>,
     luisAPITPS?: number,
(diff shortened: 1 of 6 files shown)


boydc2014:
I know the issue is described in #855, but it would be good to summarize what this PR actually did for better review transpancy

Issue #887, opened 2020-07-10
bf qnamaker:kb:export doesn't respect HTTPS_PROXY


felixvanleeuwen:
## Versions
What CLI version are you using.
botframework-cli@4.10.0-preview.143194
What Nodejs version are you using
12.18.1
What command-line interpreters are you using
Powershell 6
What OS are you using
Windows 10

## Describe the bug
bf qnamaker:kb:export doesn't respect HTTPS_PROXY.
Wireshark shows direct connection attempts to 13.69.68.23.

## To Reproduce
1. Get behind a corporate proxy;
2. Configure HTTPS_PROXY and HTTP_PROXY in system variables;
3. Run `bf qnamaker:kb:export`;
4. Error in screenshot after about a minute;
5. Output is empty

## Expected behavior
Expected output to be filled with QnA model.

## Screenshots
![image](https://user-images.githubusercontent.com/64589372/87141332-31d93e00-c2a3-11ea-8ea3-f952f7cfff67.png)

## Additional context
[bug]

munozemilio:
Hello @felixvanleeuwen, thanks for reporting this. I will take it as port of our next release cycle.

daveta:
@munozemilio : Is this a part of R10?  Could you label it accordingly?
Thanks, -Dave

munozemilio:
closed by #899

PR #899, opened 2020-07-15
Updated serviceBase.js to fix the HTTPS_PROXY

IanSeinstraDSW:
Changed the temporary variable name to nodeFetch because the global.fetch wasn't overwriting the variable.
This should fix issue 887: 
https://github.com/microsoft/botframework-cli/issues/887


EricDahlvang:
@munozemilio  Would you please take a look at this PR?

Issue #888, opened 2020-07-13
Improve the Lu intents CRUD performance


lei9444:
## Versions
master branch

## Describe the bug
It will take 2 seconds when updating the lu intents in an 1k lines lu file.
If we can reduce the waiting time. It will be helpful for composer' UI render performance

PR #901, opened 2020-07-16
Improve the CRUD performance of LU sections

Danieladu:
Fixes: #888 
Improve the performance for adding/updating/deleting/inserting section into LUResource.

## What this pr did:
1. Replace parseTree with "range" in section
2. Only parse corresponding lu content to avoid the whole file parsing
3. Adjust section and error range

## what will do
- [x] Pass all tests
- [x] Provide performance-enhancing data reports

## what should be discussed
Currently, LU has no static checker. So, if someone updates a section, the diagnostic of the original section should be removed.
The issue is, it is hard to locate the original errors, for many of them do not contain the range info.
Welcome to propose a solution.

## performance improvement

  | Before | After
-- | -- | --
Parse whole file | 600ms | 600ms
CRUD time | 400ms(With builtin cache) | 5ms


packages/lu/src/parser/lufile/luParser.js
```diff
@@ -14,22 +14,42 @@ const LUErrorListener = require('./luErrorListener');
...
 class LUParser {
+
     /**
-     * @param {string} text
+     * 
+     * @param {string} text 
+     * @param {LUResource} luResource 
      */
-    static parse(text) {
+    static parseWithRef(text, luResource) {
```
feich-ms, line 28: >static parseWithRef(text, luResource) { [](start = 4, length = 39)

what about return luResource for text is empty?

packages/lu/src/parser/lufile/luParser.js
```diff
@@ -50,12 +70,21 @@ class LUParser {
             } else {
                 nestedIntentSections.forEach(section => {
                     let emptyIntentSection = new SimpleIntentSection();
-                    emptyIntentSection.ParseTree = section.ParseTree.nestedIntentNameLine();
                     emptyIntentSection.Name = section.Name;
+                    emptyIntentSection.Id = `${emptyIntentSection.SectionType}_${emptyIntentSection.Name}`
+                    
+                    // get the end character index
+                    const firstLine = content.split(/\r?\n/)[0];
+                    let endCharacter = section.Name.length + 2;
```
feich-ms, line 78: >let endCharacter = section.Name.length + 2; [](start = 20, length = 43)

this may not correct as intent name can be '# name' or '## name' or '### name'

packages/lu/src/parser/lufile/parseFileContents.js
```diff
@@ -731,7 +731,6 @@ const parseAndHandleNestedIntentSection = function (luResource, enableMergeInten
         sections.forEach(section => {
             if (enableMergeIntents) {
                 let mergedIntentSection = section.SimpleIntentSections[0];
-                mergedIntentSection.ParseTree = section.ParseTree;
```
feich-ms, line 734: >mergedIntentSection.ParseTree = section.ParseTree; [](start = 16, length = 50)

the mergedIntentSection's range is not set. It is not the first simpleIntentSection, instead it merged other utterances and entities in the same nested intent.

packages/lu/src/parser/lufile/parseFileContents.js
```diff
@@ -1140,28 +1139,28 @@ const parseAndHandleEntityV2 = function (parsedContent, luResource, log, locale)
...
-                        handleComposite(parsedContent, entityName,`[${candidateChildren.join(',')}]`, entityRoles, entity.ParseTree.newEntityDefinition().newEntityLine(), false, entity.Type !== undefined);
+                        handleComposite(parsedContent, entityName,`[${candidateChildren.join(',')}]`, entityRoles, entity.Range, false, entity.Type !== undefined);
                         break;
                     case EntityTypeEnum.LIST:
-                        handleClosedList(parsedContent, entityName, entity.ListBody.map(item => item.trim()), entityRoles, entity.ParseTree.newEntityDefinition().newEntityLine());
+                        handleClosedList(parsedContent, entityName, entity.ListBody.map(item => item.trim()), entityRoles, entity.Range);
                         break;
                     case EntityTypeEnum.PATTERNANY:
-                        handlePatternAny(parsedContent, entityName, entityRoles, entity.ParseTree.newEntityDefinition().newEntityLine());
+                        handlePatternAny(parsedContent, entityName, entityRoles);
```
feich-ms, line 1148: > handlePatternAny(parsedContent, entityName, entityRoles); [](start = 23, length = 58)

no range parameter here?

packages/lu/src/parser/lufile/luParser.js
Danieladu: This method just return the `luResource` which the "text" contains. But not the whole original luResource.

packages/lu/src/parser/lufile/luParser.js
Danieladu: Yes, `let endCharacter = section.Name.length + 2;` is just a fallback solution, the real assign is 
```
const firstLine = content.split(/\r?\n/)[0]
if (firstLine.includes(section.Name)) {
   endCharacter = firstLine.length;
}
```

Currently, it is the only way to get the section name line.

packages/lu/src/parser/lufile/parseFileContents.js
Danieladu: hmm. Sorry, I'm not familiar with this merge part. Could I just change it to  `mergedIntentSection.Range= section.Range`?

packages/lu/src/parser/lufile/parseFileContents.js
Danieladu: #Fixed

packages/lu/src/parser/lufile/sectionOperator.js
```diff
@@ -14,60 +16,239 @@ class SectionOperator {
...
     sectionContent = helpers.sanitizeNewLines(sectionContent);
-    const section = this.Luresource.Sections.find(u => u.Id === id);
-    if (!section) {
+    const sectionIndex = this.Luresource.Sections.findIndex(u => u.Id === id);
+    if (sectionIndex < 0) {
+      return this.Luresource;
+    }
+
+    const oldSection = this.Luresource.Sections[sectionIndex];
+    const newResource = luParser.parseWithRef(sectionContent, this.Luresourcehis);
```
feich-ms, line 53: >Luresourcehis [](start = 67, length = 13)

naming is not right

packages/lu/src/parser/lufile/parseFileContents.js
feich-ms: yes, let's do this

---
In reply to: [456200240](https://github.com/microsoft/botframework-cli/pull/901#discussion_r456200240) [](ancestors = 456200240)

packages/lu/src/parser/lufile/nestedIntentSection.js
```diff
@@ -22,15 +25,18 @@ class NestedIntentSection {
...
 
     ExtractName(parseTree) {
         return parseTree.nestedIntentNameLine().nestedIntentName().getText().trim();
     }
 
     ExtractBody(parseTree, content) {
-        const startLine = parseTree.start.line - 1
-        const stopLine = parseTree.stop.line - 1
+        const startLine = parseTree.start.line;
+        const stopLine = parseTree.stop.line;
```
feich-ms, line 39: why change this? The body here is excluding nested intent name. In simpleIntentSection, it is not changed.

feich-ms:
@lei9444 please verify the functionality and efficiency of this improvments in composer side.

PR #903, opened 2020-07-16
Able to add any extension as a CLI pattern

chrimc62:
This allows adding .uischema files and other extensions to better support testing and ad hoc use.
Checked-in generated C# support files even though projects are just for
testing.


packages/dialog/test/commands/dialog/projects/project5/obj/Debug/netcoreapp2.1/project5.AssemblyInfoInputs.cache
```diff
@@ -0,0 +1 @@
+ef6dea4726012fb20aae4e9fab90c2ca7a8d450b
```
munozemilio, line 1: Is this intended?

PR #905, opened 2020-07-20
Continue improving the LU section CRUD's performance

Danieladu:
Fixed:  # 906
Continue improving the LU section CRUD's performance.
## What the previous PR  #901 has done:
- Replace property `parseTree` with `range` in section
- Only parse corresponding lu content to avoid the whole file parsing
- Use `cloneDeep` and `cloneDeepWith` to keep the input resource Immutable
-  Adjust section and error range

## The effect of the previous PR  
- Avoid the whole file parsing to improve performance.

## What this PR did:
- Remove unused code
- Replace `cloneDeep` and `cloneDeepWith` with `JSON.parse(JSON.stringify(luresource));` to improve the clone performance.

## The effect of this PR  
- Reduce the clone time of Section CRUD operator.
Use `cloneDeep` and `cloneDeepWith` : 50ms
Just use JSON serialization and Deserialization: < 1ms

## What next would do:
- Adjust grammar file of LU parser to improve the parsing performance.

## The effect of the next PR  
- Improve the performance of parsing a single LU file


boydc2014:
@zhixzhan can you help check this

boydc2014:
@Danieladu also, to avoid confusing in PR review, please list more context, what previous you did to improve perf, and what's this one for, and any other further plans to improve performance.

zhixzhan:
> @zhixzhan can you help check this

Sure, my another question is 
```
const newResource = new sectionOperator(resource).updateSection(targetSection.Id, targetSectionContent);
```

this methods modify on parameter `resource`, but also return `newResource` that could probably  make little confuse, and even cause error when parameter is defined as `immutable`. 

I suggest do not modify parameter `resource`, which make API implementation is `Pure Function` has no side effect.

Danieladu:
> > @zhixzhan can you help check this
> 
> Sure, my another question is
> 
> ```
> const newResource = new sectionOperator(resource).updateSection(targetSection.Id, targetSectionContent);
> ```
> 
> this methods modify on parameter `resource`, but also return `newResource` that could probably make little confuse, and even cause error when parameter is defined as `immutable`.
> 
> I suggest do not modify parameter `resource`, which make API implementation is `Pure Function` has no side effect.

Yes, the input luresource would not be modified. But each CRUD operator is Iterative.
For example:
```
var luresource = xxx;
const operator = new SectionOperator(luresource);
const newresource1 = operator.updateSection(yyy);
const newresource2 = operator.updateSection(zzz);
```

The operator: `const newSection2 = operator.updateSection(zzz);` is based on the resource `newresource1 `

PR #908, opened 2020-07-20
entity definition of missing entity type will be included in simpleIntentSection

### LINKED ISSUE #907: LU parser update  body not correct.
## Browser

- [ ] Electron distribution
- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

## OS

- [ ] macOS
- [ ] Windows
- [ ] Ubuntu

## Screenshots
image

feich-ms:
fix #907. Before the fix, entity line of '$a' after a intent will not be included into simpleIntentSection. After the fix, it will be included as body of the simpleIntentSection and fill an error in parse Errors. This will help composer lu editor.


feich-ms:
Close this one since it is already merged into PR #910 as #910 is also optimizing lu parser.

Issue #861, opened 2020-06-17
Publish preview builds to NPM


munozemilio:
**Is your feature request related to a problem? Please describe.**
Myget adds friction to the customer when they want to try new features or available fixes for future releases

**Describe the solution you'd like**
Have 3 release streams published to NPM:
- RC
- Latest
- Next

Please refer to #864 
[enhancement]

compulim:
❤ this feature! FYI, `botframework-directlinejs` is on this one already. `botframework-webchat` is semi on this one, we [publish bundle to GitHub](https://github.com/microsoft/BotFramework-WebChat/releases/daily).

I think every package should have a good version number to aid the devs know how cutting edge they are on:

- Devs can easily know what version they are running
   - This will aid us on supporting bugs: may be the bug has already resolved but they are not cutting edge enough. Or maybe the fix doesn't help them.
- Devs can easily find out if a specific version has a specific PR or not
   - You can look at [how Web Chat done this](https://github.com/microsoft/BotFramework-WebChat/releases/daily)

PR #911, opened 2020-07-21
Build yaml files refactor

munozemilio:


munozemilio:
FIxes #861

build/botframework-cli-build-test.yml
```diff
@@ -0,0 +1,73 @@
+steps:
```
BruceHaley, line 1: It would be nice if this template file was named with "-steps" at the end. That lets us more easily distinguish between root yaml files and template yaml files. That is the standard we are using elsewhere. For example: https://github.com/microsoft/botbuilder-dotnet/blob/master/build/yaml/ci-build-steps.yml

PR #925, opened 2020-08-06
Fix loadContents not working issue when loading both non-empty and empty files in luis/qnamaker:build

### LINKED ISSUE #924: loadContents function will fail when empty and non-empty lu or qna files list are provided
## Versions
bf-cli master branch
Composer reported issue since they have empty lu or qna files in many scenarios. 

## Describe the bug
When users call loadContents function in builder.ts of lubuild or qnabuild, it will not load files and create dialogs assets correctly when parsing empty files. The root cause it that luMerger will throw exception if we don't filter those empty files. The fix is straightforward, just filter empty files before calling luMerger and avoid creating recognizers for those empty files.

## To Reproduce
1. use luis build or qnamaker build cli
2. use one non-empty and one empty files as input
3. see the build results, you will find the non-empty file is not published

## Expected behavior
loadContents can handle empty files correctly.

## Additional context

[bug]

feich-ms:
Fix #924 to make loadContents function load files correctly when both non-empty and empty files are there at the same time.

Tests are added.

Composer reported issue since they have empty lu or qna files in many scenarios.

### DIFF
--- a/packages/lu/src/parser/lu/luMerger.js
+++ b/packages/lu/src/parser/lu/luMerger.js
@@ -547,7 +547,7 @@ const parseLuFile = async function(luOb, log, luis_culture) {
     let parsedContent = ''
     if (!luOb.content) {
         let error = BuildDiagnostic({ message: `Cannot parse empty ${luOb.id}. Please add content to the file or remove it.` })
-        throw(new exception(retCode.errorCode.EMPTY_CONTENT, error.toString()));
+        throw(new exception(retCode.errorCode.INVALID_INPUT_FILE, error.toString()));
     } 
     try {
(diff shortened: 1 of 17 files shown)


feich-ms:
@munozemilio this PR is to fix an issue reported by composer team. They have some scenarios that lu or qna files are empty but still need to be passed into lu or qna build. As you know, luMerger will throw exception for empty files, before the fix, the exception will block other non-empty files publishing, so I made a change in luis/qna build to filter empty files before calling luMerger. This has no impact on any other components but really helps composer team. I added corresponding tests there, I merged this first since it is blocking issue for composer team. Please let me know if you have any concerns.

PR #931, opened 2020-08-11
Removing Dispatcher from master

munozemilio:


vishwacsena:
@boydc2014 @feich-ms FYI. This will be moved off master to a preview branch.

PR #932, opened 2020-08-12
Change dialog:merge bundling

chrimc62:
Implement bundling so that it keeps the relationship to the original source of external schema.  This makes schemas easier to understand and simplifies how Composer deals with schema.


packages/dialog/src/library/schemaMerger.ts
```diff
@@ -1347,6 +1347,126 @@ export default class SchemaMerger {
         }
     }
 
+    // Split a $ref into path, pointer and name for definition
```
munozemilio, line 1350: Unit tests for this new block not needed?

chrimc62: Added a bunch of new unit tests with more complex schema that exercise this.

---
In reply to: [469073030](https://github.com/microsoft/botframework-cli/pull/932#discussion_r469073030) [](ancestors = 469073030)

cleemullins:
@munozemilio Please merge when ready. Do we need to cherrypick to R10?

PR #935, opened 2020-08-13
Expose import resolver interface for cross-train/luis:build/qnamaker:build

### LINKED ISSUE #934: CustomizedImportResolver in LU build/crossTrain
Composer use are use those modules do publishing:
```
const crossTrainer = require('@bfcomposer/bf-lu/lib/parser/cross-train/crossTrainer.js');
const luBuild = require('@bfcomposer/bf-lu/lib/parser/lubuild/builder.js');
const LuisBuilder = require('@bfcomposer/bf-lu/lib/parser/luis/luisBuilder');
```
3 methods need support customized import resolver: 
- `loadContents` in `lubuild.Builder`  and `qnabuild.Builder`
- `crossTrain` in `crossTrainer`

feich-ms:
fix #934 to expose import resolver interface for cross-train/luis:build/qnamaker:build so that composer can customize their file resolving system based on their import system.


packages/lu/src/parser/cross-train/crossTrainer.js
```diff
@@ -28,19 +28,20 @@ module.exports = {
    * @param {any[]} luContents the lu content array whose element includes path and content
    * @param {any[]} qnaContents the qna content array whose element includes path and content
    * @param {any} crossTrainConfig cross train json config
+   * @param {any} fileResolver file resolver when resolving import files
    * @returns {Map<string, LUResource>} map of file id and luResource
    * @throws {exception} throws errors
    */
-  crossTrain: async function (luContents, qnaContents, crossTrainConfig) {
+  crossTrain: async function (luContents, qnaContents, crossTrainConfig, fileResolver) {
```
boydc2014, line 35: can we call this importResolver to align with lg concepts?

boydc2014: and importResolver is more generic than "fileResolver" since not always every resource comes from file

feich-ms: fixed

PR #944, opened 2020-08-18
Sibling $ref required removed from definition.

chrimc62:
The new bundler exposed an issue in our component schema.  When using Activity there would be a $ref sibling required.  This conflicted with the required in Activity.  This is supported in JSON Schema 8, but fixed this for now by removing the definition required.  This makes schema more readable and may be needed for Composer release.


.vscode/launch.json
```diff
@@ -194,12 +194,14 @@
...
+                "libraries/**/*.schema",
+                "tests/**/*.schema",
                 "-o",
-                "schemas/app.schema"
+                "tests/tests.schema",
+                "--verbose"
             ],
             "internalConsoleOptions": "openOnSessionStart",
-            "cwd": "${workspaceFolder}/packages/dialog/test/commands/dialog"
+            "cwd": "${workspaceFolder}/../botbuilder-dotnet/"
```
munozemilio, line 204: this may not work for others

PR #951, opened 2020-08-21
Update BF luis:application:create - removed invalid options from example

WashingtonKayaker:


munozemilio:
This is autogenerated from command files.

PR #955, opened 2020-08-24
fix update kb issue in importUrlReference and importFileReference APIs

### LINKED ISSUE #953: importUrlReference in qna builder lib should update or create kb correctly
## Versions
latest version 4.10.1

## Describe the bug
importUrlReference or importFileReference should update or create kb correctly if kb is not deleted in some cases.

## To Reproduce
1. call importUrlReference or importFileReference function and cancel the call before deleting is executed.
2. You will see the kb is created and not deleted
3. call importUrlReference or importFileReference function with the same kb name
4. it will throw errors like `{\n "error": {\n "code": "BadArgument",\n "message": "Invalid input. See details.",\n "details": [\n {\n "code": "ValidationFailure",\n "message": "'Qn A List' should not be empty.",\n "target": "QnAList"\n }\n ]\n }\n}'`

## Expected behavior
No errors are thrown and the kb should be updated or created

## Additional context

[bug]

feich-ms:
fix #953. Before the fix, it will fail when importing url or file reference into an existing kb since replacing kb with empty content is not allowed. To unblock the importing, we have to delete the kb and create a new one. This will be only applied for importUrlReference and importFileReference APIs in qna builder.

### DIFF
--- a/packages/lu/src/parser/qnabuild/builder.ts
+++ b/packages/lu/src/parser/qnabuild/builder.ts
@@ -313,15 +313,13 @@ export class Builder {
       }
     }
 
-    // compare models to update the model if a match found
-    // otherwise create a new kb
+    // delete the kb if it already exists
     if (kbId !== '') {
-      // To see if need update the model
-      await this.updateUrlKB(qnaBuildCore, url, kbId)
-    } else {
-      // create a new kb
-      kbId = await this.createUrlKB(qnaBuildCore, url, kbName)
+      await qnaBuildCore.deleteKB(kbId)
     }
+
+    // create a new kb
+    kbId = await this.createUrlKB(qnaBuildCore, url, kbName)
 
     const kbJson = await qnaBuildCore.exportKB(kbId, 'Test')
     const kb = new KB(kbJson)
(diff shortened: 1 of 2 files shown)


feich-ms:
@munozemilio @vishwacsena this is bug fix for importUrlReference and importFileReference APIs that composer are consuming. The root cause of this issue is that, before the fix, importUrlReference or importFileReference has update kb logic in which I use replace api to clean the kb. But actually it's not allowed to replace an existing kb with empty content. So it will always fail when composer call these APIs to import url or file reference into an existing kb that is not deleted in last call due to some interruptions. Currently composer is using uuid as kb name when importing url or file reference to avoid importing content into any existing kb. This can solve the problem, but may leave some unexpected kbs in user's qna portal if the import is interrupted which means the url or file reference kb is created but not deleted. After the fix, composer can use a meaningful kb name to import the url or file reference. Even if the import is interrupted and kb is not deleted, the import will still work in next round and users will feel more friendly when they see the kb with specific name in their qna service portal.

If possible, we hope this can catch next rc release of cli, but of course not must-have.

PR #961, opened 2020-09-02
bf-lu: support new link reference format

### LINKED ISSUE #917: LU format link reference additions
LU format today only supports `[description](link)` notation for imports as well as resolving references. We should expand this to support the following markdown notation

```markdown
[import][1]

# intent 1
- [reference][2]

[1]:foo.lu
[2]:bar.lu#intent2
```

feich-ms:
fix #917 to support new import format in lu file like
```markdown
[import][1]

# intent 1
- [reference][2]

[1]:foo.lu
[2]:bar.lu#intent2
```


vishwacsena (changes requested):
Can you add negative tests to verify intended behavior for 
- invalid syntax (e.g. [foo[1]] or missing closing parenthesis etc)
- unable to resolve reference

Can you also add a test to verify that the reference can be to a URL? 

[foo][1]
[1]: https://blobstore.azurewebsites.net/1.lu#myTrigger

feich-ms:
@vishwacsena 
1. negetive tests are added in this commit https://github.com/microsoft/botframework-cli/pull/961/commits/957395a5bc09c977c2354bd98788fb0ec58d8e34. For invalid import syntax in utterances, it's not easy to validate since utterance can tolerate all kinds of inputs and we can't distinguish it's invalid import or just utterance itself, so I didn't add tests for utterance import syntax error(Tests for invalid reference file or URL in utterance are added).
2. test for URL reference is added in this commit https://github.com/microsoft/botframework-cli/pull/961/commits/58d7be39030fda2017fe162b893538edabf6c627

vishwacsena (approved):

PR #962, opened 2020-09-02
add version in luis recognizer and settings file and adjust deleteOldVersion interface

feich-ms:
Per requirememts of version control, add version property in luis recognizer and settings file to enable version control. And also adjust deleteOldVersion to keepVersions  in luis builder interface. The keepVersions parameter is number of kept luis application versions when updating. Composer can set it to a reasonable value(such as 5 or 10 or even 100) when consuming this keepVersions interface. For luis build CLI, no interface changes. If the deleteOldVersion is set true, that means keeping only one latest version and delete all other versions, if deleteOldVersion is not set or set to false, the luis build will keep all versons up to 100(LUIS limit).


packages/lu/src/parser/lubuild/builder.ts
```diff
@@ -159,7 +161,7 @@ export class Builder {
     botName: string,
     suffix: string,
     fallbackLocale: string,
-    deleteOldVersion: boolean,
+    keepVersions: number,
```
boydc2014, line 164: Overall looks good, one nit is this keepVersions don't seems to be a good name, how about "keepVersionCount", or something more common?

feich-ms: changed it to "keptVersionCount"

tomlm (approved):
:shipit:

PR #978, opened 2020-09-16
Update CLI README.md

WashingtonKayaker:
The `qnamaker:build` command is missing the `qnaConfig` option.


munozemilio:
This readme is autogenerated from the plugins. The change has to be done at the command file level

PR #990, opened 2020-09-25
Refactor luis build interface to decouple dialog gen from build

### LINKED ISSUE #985: Decouple declarative dialog generation from luis:build and qna:build
Composer is leveraging bf luis:build and bf qna:build to do luis publishing and qna publishing. 

Besides publishing, luis:build and qna:build do another work that is generating a bunch of .lu.dialog, .qna.dialog, .lu.qna.dialog, which is 
* built into it's core interface and run together with this build command. 
* it's using a fixed set of template to generate

The direction of Composer is heading is Composer would like to take control of those declarative dialog generation part which will allow user to author those dialogs to enable\disable certain features. And also make it friendly to CI/CD environment, so that the composer bot can be integrated to CI/CD pipeline easily. 

The ask to cli, is that we should decouple the declarative dialog generation out from it's core capability or at least make it optional.

feich-ms:
fix #985 in cli and https://github.com/microsoft/BotFramework-Composer/issues/4259 in Composer


packages/cli/README.md
```diff
@@ -771,6 +771,7 @@ OPTIONS
 
   --isStaging                      Publishes luis application to staging slot if set. Default to production slot
 
+  --genSettingsOnly                [default: false]Indicates only write out settings to out folder. Only valid if --out is set
```
boydc2014, line 774: I'm not sure we want to introduce a new settings field, we should leverage the "--dialog" option. 

If no "--dialog" is specified, we don't generate .dialog.
If "--dialog" is specified, we generate --dialog as specified value, either "MultiLanguage" or "Cross-Train", this is purely for backward compatible

feich-ms: fixed.

lei9444:
@feich-ms I have a questions about the empty lu file.
In this PR, if I use an empty file and run loadContents, The result is like 
image

but in version 4.10.0, the result is empty
image

I am not sure which result is correct?

packages/luis/src/commands/luis/build.ts
```diff
@@ -125,22 +117,15 @@ export default class LuisBuild extends Command {
 
         // load lu contents from lu files
         // load existing recognizers, multiRecogniers and settings or create default ones
-        const loadedResources = await builder.loadContents(files, defaultCulture, suffix, region, schema)
-        luContents = loadedResources.luContents
-        recognizers = loadedResources.recognizers
-        multiRecognizers = loadedResources.multiRecognizers
-        settings = loadedResources.settings
-        crosstrainedRecognizers = loadedResources.crosstrainedRecognizers
+        luContents = await builder.loadContents(files, {
```
lei9444, line 120: @feich-ms I have a questions about the empty lu file.
In this PR, if I use an empty file and run loadContents, The result is like
image

but in version 4.10.0, the result is empty
image

I am not sure which result is correct?

feich-ms: @lei9444, that's ok. In this PR, I added the lu content even for empty lu file and this will be useful when generating crosstrained recognizer because we still need to generate top-level crosstrained recognizer for empty lu file.

boydc2014 (approved):

feich-ms:
@munozemilio, Composer has validated the functionality of this PR, so I merged it to avoid resolving conflicts if they are moved out from bf-lu. Please let me know if there are any concerns.

PR #1011, opened 2020-10-10
Add down sampling cli to beta branch

### LINKED ISSUE #991: Add cli for @microsoft/bf-dispatcher
Now composer use @microsoft/bf-dispatcher as a library in server to do down sampling.

Now composer use bootstramp sampling to balance the utterances' number ratio, and use reservoir sampling to down size the utterances.

To take control of the dialog generation, we want to unify the cli tools. But the bf-dispatcher does not support cli now.
what we want maybe like: do down sample for lu file, then output a new lu file with the result.

feich-ms:
fix #991: add down sampling core lib and cli to beta branch so that composer can consume the lib directly and users can consume the standalone cli tool to do down sampling on their lu files.


feich-ms:
@munozemilio, this feature is targetting R11 for composer, please help with the npm package release if you have time. The failure errors of CI comes from orchestrator that I didn't touch and have no knowledge to resolve. The tests for my features are all passed. I think this package is ready out for consuming and trying by composer team. Thank you very much.

feich-ms:
I merged this PR first so that the sampler package can be picked to release to npm service. I will let composer to consume it once it's there. Any fixes or optimizations can be taken later based on feedback.

PR #1022, opened 2020-10-16
fix exit code not thrown issue when cli running into exception

### LINKED ISSUE #1005: bf command line tools silently fail
## Versions
4.10, 4.11

## Describe the bug
When you use bf cli as part of a script when it fails it doesn't set the process error code which means that it silently fails.  Our lu files stopped building because of some internal error and for the last month we have not updated our models because the silent failure.  

## To Reproduce
create a cmd file like this:
```
call bf luis:build --foo
if errorlevel 1 @echo error 
if not errorlevel 1 @echo success
```

## Expected behavior
It should say "error" and it says "success"

[bug]

feich-ms:
fix #1005

### DIFF
--- a/packages/chatdown/test/commands/chatdown.test.ts
+++ b/packages/chatdown/test/commands/chatdown.test.ts
@@ -5,6 +5,7 @@ describe('chatdown', () => {
     test
     .stdout()
     .command(['chatdown', '--help'])
+    .exit(1)
     .it('should print the help contents when --help is passed as an argument', ctx => {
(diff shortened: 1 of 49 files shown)


boydc2014 (approved):

feich-ms:
@munozemilio hi, this is resolving a bug raised by Tom #1005, the fix is just one line code 'this.exit(1)' in command.ts, all the other changes are unit tests adjustments. please let me know if you have any concerns about these changes.

PR #1036, opened 2020-10-26
Replacing node-fetch with axios to get proxy detection

### LINKED ISSUE #968: Proxy settings are not universally respected
## Versions
**What CLI version are you using.**
@microsoft/botframework-cli/4.11.0-dev.20200907.c063415
**What Nodejs version are you using**
win32-x64 node-v12.18.1
**What command-line interpreters are you using**
Powershell 6 (x64)
**What OS are you using**
Windows 10

## Describe the bug
**Give a clear and concise description of what the bug is.**
Before each BF CLI command node tries to access registry.npmjs.org without being aware of proxy settings.
This retries for about a minute and then the command actually goes through in about 2-3 seconds.
We have a script that runs about 30-40 of these commands and it takes about 45 minutes to run it.
When a temporary exception is made for this address in our firewall the script takes about 3-4 minutes to run.

## To Reproduce
**Steps to reproduce the behavior:**
1. Get behind a corporate proxy;
2. Get behind a corporate firewall;
3. Configure Node to use the proxy;
4. Set environment variables to use the proxy;
5. Run any of the following commands:
- `bf luis:version:export`
- `bf luis:version:import`
- `bf luis:train:run`
- `bf luis:application:publish`
- `bf qnamaker:kb:export`
- `bf qnamaker:kb:replace`
- `bf luis:version:rename`
- `bf luis:convert`
- `bf luis:version:list`
6. Wait for about a minute for the command to complete;
7. That was the bug.

## Expected behavior
**Give a clear and concise description of what you expected to happen.**
Every request that the BF CLI (and its dependencies) makes respects proxy settings, whether environment or Node.

[bug]

munozemilio:
Fixes #968

### DIFF
--- a/packages/qnamaker/utils/index.js
+++ b/packages/qnamaker/utils/index.js
@@ -43,11 +43,24 @@ module.exports = async function qnamaker(config, serviceManifest, args, requestB
     const service = new api[identifier]();
     const response = await service[operation.name](args, (requestBodyDataModel || requestBody));
     //console.log(' requestBodyDataModel '+requestBodyDataModel +' entityType '+operation.entityType+' requestBody '+ requestBody)
-    const text = await response.text();
+
     try {
-        return JSON.parse(text);
+        return response.data;
     }
-    catch (e) {
-        return text;
+    catch (error) {
+        if (error.response) {
+            // The request was made and the server responded with a status code
+            // that falls out of the range of 2xx
+            return error.response.data;
+
+          } else if (error.request) {
+            // The request was made but no response was received
+            // `error.request` is an instance of XMLHttpRequest in the browser and an instance of
+            // http.ClientRequest in node.js
+            return error.request;
+          } else {
+            // Something happened in setting up the request that triggered an Error
+            return error.message;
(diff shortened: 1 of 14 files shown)


packages/qnamaker/utils/index.js
```diff
@@ -43,11 +43,24 @@ module.exports = async function qnamaker(config, serviceManifest, args, requestB
     const service = new api[identifier]();
     const response = await service[operation.name](args, (requestBodyDataModel || requestBody));
```
feich-ms, line 44: >const response = await service[operation.name](args, (requestBodyDataModel || requestBody)); [](start = 4, length = 92)

Maybe this line can be moved into try-catch block? I see in http-request.ts, 'await axios(config)' is  inside try-catch. But here, 'await service' seems equivalent to 'await axios(config)', but not in try-catch. Not sure this impacts or not. Feel free to close it if has not impact. Other changes look good to me.

feich-ms (approved):
One minor comment. Looks good to me.

PR #1038, opened 2020-10-27
Optimize empty file validation and import resolver corner cases in crossTrianer

feich-ms:
Two changes required from composer
1. Before the change, the empty lu file validation is based on condition that content !== '' which is not exactly correct for some corner cases such as only comments or model informations are in lu file. After the changes, the empty file validation is based on non model info section count which means the file is not empty only if the sections(except model info section) length greater than 0. Without this fix, the crosstrained lu results may not be expected.
2. Add .lu or .qna extension for lu and qna object id when resolving imports if they don't have extension since composer has a customized resolver that requires the id with extension.

### DIFF
--- a/packages/lu/src/parser/cross-train/crossTrainer.js
+++ b/packages/lu/src/parser/cross-train/crossTrainer.js
@@ -293,7 +293,8 @@ const mergeInterruptionIntent = function (fromUtterances, toResource, intentName
 
       // add section here
       // not add the interruption intent if original file is empty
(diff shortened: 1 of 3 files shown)


feich-ms:
Hi @munozemilio, this PR is to fix a potential bug. Composer will consume this fix. I saw the first cli rc release is already out, so hope it can catch next rc release if possible. Could you please help to cherry-pick it to next rc release? Thank you very much.

boydc2014 (approved):

PR #1039, opened 2020-10-28
[Cherry-pick to 4.11] Optimize empty file validation and import resolver in crossTrainer core lib

feich-ms:
Cherry-pick PR #1038 to 4.11 branch. This fix is required by composer cross-train functionality. Without this, cross-train will fail in composer.


feich-ms:
close this one since it accidently targetted main branch and on way to retarget 4.11 branch.

PR #1042, opened 2020-10-30
Fix a regression on multiturn extraction functionality of qna url or file import api

feich-ms:
Composer team reported a regression issue https://github.com/microsoft/BotFramework-Composer/issues/4590 on the multirun extraction functionality of url and file import api in bf-lu library. This was caused by a bad [merge ](https://github.com/microsoft/botframework-cli/pull/990/files#diff-a0253f24b1d7f993c390269c8b172cea0d59a480556f55853ad675d6a5f1f78bL303)before. To avoid such regression in the future, more specific tests are added in this PR.

### DIFF
--- a/packages/lu/src/parser/qnabuild/builder.ts
+++ b/packages/lu/src/parser/qnabuild/builder.ts
@@ -301,7 +301,9 @@ export class Builder {
     url: string,
     subscriptionkey: string,
     endpoint: string,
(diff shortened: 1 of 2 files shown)


boydc2014 (approved):

boydc2014:
Can we add some test case to avoid regression like this?

PR #1045, opened 2020-10-30
Make luis application name comparison case insensitve in luis build

### LINKED ISSUE #1044: Luis:build should compare luis application name case insensitively
## Versions
4.11

## Describe the bug
Luis build chooses to create or update an application based on comparing current app name with all the existing app names. Currently the comparison is case sensitve, so when an app name is only case different with the one of exising app names, the luis build will try to create a new app with that name. However, app name is case insensitive in luis service, so luis service will reject the creation of same app name that only case different and throw an error. This is not expected.

## To Reproduce
1. Use bf luis:build command to publish a lu file with botName set to xxx
2. Use same command to update the luis model but with botName set to XXX
3. You will see exception thrown

## Expected behavior
No errors are thrown out

## Screenshots
image

## Additional context

[bug]

feich-ms:
Fix #1044 and https://github.com/microsoft/BotFramework-Composer/issues/4627. The fix is to align with luis service to make application name case insensitive in luis build.

App name in luis service is case insensitive, for example, 'my_app_name' and 'My_App_Name' will be treated as same app name. But in current luis build, the name comparison is case sensitive. This PR is trying to make the app name comparison case insensitive in luis build to avoid such inconsistence. 

BTW, qna kb name is case sensitive in qnamaker service, so no need to change the comparison logic in qna build.

### DIFF
--- a/packages/lu/src/parser/lubuild/builder.ts
+++ b/packages/lu/src/parser/lubuild/builder.ts
@@ -232,7 +232,7 @@ export class Builder {
               // find if there is a matched name with current app under current authoring key
               if (!recognizer.getAppId()) {
                 for (let app of apps) {
-                  if (app.name === currentApp.name) {
+                  if (app.name.toLowerCase() === currentApp.name.toLowerCase()) {
                     recognizer.setAppId(app.id)
                     break
                   }
--- a/packages/luis/test/commands/luis/build.test.ts
+++ b/packages/luis/test/commands/luis/build.test.ts
@@ -318,6 +318,38 @@ describe('luis:build not update application if no changes', () => {
     })
 })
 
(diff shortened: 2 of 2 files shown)


boydc2014 (approved):

feich-ms:
Hi @munozemilio, this is a fix for luis application name case sensitive issue in luis build which was reported by composer team. The root cause is that luis build is not aligned with luis service in application name case sensitivity. To align with luis service whose app name is case insensitive, we changed the comparison logic in luis build to make the app name comparison case insensitive too. Of course, this is just an optimization task instead of a must-have bug fix, but if we get chances to have next planned rc iteration, it will be helpful to include this by the way. Thank you.

PR #1048, opened 2020-11-04
Update build.ts

WashingtonKayaker:
Added "(required) " to descriptions for the following required options:

1. in
1. botName
1. authoringKey
1. endpoint


munozemilio:
@feich-ms could you please review?

packages/luis/src/commands/luis/build.ts
```diff
@@ -38,7 +38,7 @@ export default class LuisBuild extends Command {
     luConfig: flags.string({description: 'Path to config for lu build which can contain switches for arguments'}),
     deleteOldVersion: flags.boolean({description: 'Deletes old version of LUIS application after building new one.'}),
     log: flags.boolean({description: 'Writes out log messages to console', default: false}),
-    endpoint: flags.string({description: 'Luis authoring endpoint for publishing'}),
+    endpoint: flags.string({description: '(required) Luis authoring endpoint for publishing'}),
```
feich-ms, line 41: The endpoint has the value with priority: user specified with --endpoint > https://{region}.api.cognitive.microsoft.com(only region is specified by user) > https://westus.api.cognitive.microsoft.com(default, nothing specified). So in some cases, this may not be required and it can use default value.

feich-ms (approved):
Left a comment for endpoint, could you address that?

munozemilio (approved):

PR #1058, opened 2020-11-23
Support special chars in intent name when parsing lu with antlr

### LINKED ISSUE #1052: LU Export doesn't work for intents with special characters in the name
## Versions
LU 4.9.1

## Describe the bug
LU export doesn't work if app has intents with special characters in the name

## To Reproduce
Create an app
Add an intent, named with special characters like "'" and "á".
Export as LU

## Expected behavior
LU Export should work fine, JSON export doesn't hit this error.

Source IcM: https://portal.microsofticm.com/imp/v3/incidents/details/212216724/home

[bug]

feich-ms:
Fix #1052 to support special chars in intent name when parsing lu file with antlr.
Tests are added for both lu parser and luis convert cmd.


feich-ms:
@munozemilio could you take a review?

munozemilio (approved):

PR #1089, opened 2021-01-14
Allow intent or entities using phraseList features of same name

### LINKED ISSUE #1051: LU Export doesn't work for intents and features with the same name
## Versions
LU 4.9.1

## Describe the bug
LU export doesn't allow intent and features to have the same name

## To Reproduce
1. Create an app
2. Add an intent, named say 'abc'
3. Add a feature to the above intent with the same name ('abc')
4. Export as LU

## Expected behavior
LU Export should work fine, JSON export doesn't hit this error.

Source IcM: https://portal.microsofticm.com/imp/v3/incidents/details/212216724/home

[bug]

feich-ms:
fix #1051. This PR is loosing the validation in lu ->json converter when intents or entities use phraseList features with same name. This is allowed in LUIS portal, but blocked in bf-lu. Also added the typical test cases for these scenarios.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -209,9 +209,9 @@
       "outFiles": ["./packages/dialog/lib/**"],
       "args": [
         "dialog:merge",
-        "runbot.csproj",
+        "RunBot.csproj",
         "-o",
-        "runbot.schema",
+        "RunBot.schema",
         "--verbose"
       ],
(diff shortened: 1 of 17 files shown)


feich-ms:
Hi @munozemilio, this is the fix for issue #1051. Could you please take a look when you are convenient. Thanks.

munozemilio (approved):

PR #1095, opened 2021-01-23
Hash line ending agnostic and import exclusion bug

chrimc62:
Made hash computation line ending agnostic by normalizing to plain line feed.  On CRLF systems this may result in a different hash than before.  We always write OS.EOL, but git might change the line endings so this change was made.
Fix bug where imports were not excluded properly if they were relative paths.  As a work-around people can add !**/imported/** to their command line.
Added obj to default exclusions.


munozemilio (approved):
Should we add a test for this?

Issue #1110, opened 2021-02-08
Luis converter failed to escape special chars like brackets


feich-ms:
## Versions
Current GA version

## Describe the bug
Luis converter failed to escape special chars like brackets.

## To Reproduce
1. Download the attached luis json file [escapeCharactersInUtterances.zip](https://github.com/microsoft/botframework-cli/files/5944327/escapeCharactersInUtterances.zip)
2. Try to convert it to lu file with bf luis:convert command
3. Try to the convert the generated lu file back to luis json file
4. You will the exceptions

## Expected behavior
No exceptions

## Additional context

[bug]

PR #1111, opened 2021-02-08
Fix luis converter to enable escaping specific chars

feich-ms:
Fix https://github.com/microsoft/botframework-cli/issues/1110. With the fix, luis converter can add necessary escape characters when converting luis json file to lu file. With these escape characters added, the lu content can be converted back to luis json file successfully without any exceptions. That means such escaping corner cases between lu and luis converting will be handled well with this fix. Test cases from customers were added to verify the fix.

Previous fix https://github.com/microsoft/botframework-cli/pull/1104 and https://github.com/microsoft/botframework-cli/pull/1106 that didn't resolve the issue completely


feich-ms:
Hi @munozemilio, could you please help to check if this fix will introduce any potential issues since it's touching luis converter logic that you have more experiences.

munozemilio (approved):

Issue #1123, opened 2021-02-19
Intent only import in lu will pull unrelated contents 


feich-ms:
## Versions
4.12.0-rc0

## Describe the bug
Reported by composer here https://github.com/microsoft/BotFramework-Composer/issues/5756#issuecomment-777048352. Intent import in lu will pull unrelated contents. Consider below scenarios:

common.en-us.lu:
```
# Greeting
- hi
- hello
```

root.en-us.lu:
```
[Common](common.en-us.lu)

# MyIntent
- test 1
- test 2
```

test.en-us.lu
```
# Test
- [MyIntent](root.en-us.lu#MyIntent)
```

The generated luis file for test.en-us.lu with luis:convert command will pull everything from common.en-us.lu, which is not expected.

## To Reproduce
1. Create above lu files
2. Use `bf luis:convert -i test.en-us.lu` to convert lu to luis
3. You will see intent and utterances in common.en-us.lu are added into the generated luis json file

## Expected behavior
Only MyIntent is pulled

## Additional context

[bug]

PR #1124, opened 2021-02-19
Fix intent import in lu merger will pull unrelated contents in some corner cases

feich-ms:
Fix https://github.com/microsoft/botframework-cli/issues/1123 to resolve the issue that intent import in lu will pull unrelated contents in some corner cases.


feich-ms:
Hi @munozemilio, this bug was just reported by composer users recently. In some corner situations, intent import will pull some unrelated contents into current luis file due to some bug in our lu merger. I made a fix here. If there is another rc1 release for bf-cli, I think it would be helpful to cherry-pick this in. Thanks.

munozemilio (approved):

PR #1129, opened 2021-02-23
Fix bug where luis:convert would incorrectly generate an error message

### LINKED ISSUE #1128: luis:convert generates incorrect error message about hierarchical labels
## Describe the bug
When converting a complex set of LU files from form generation with hierarchical entities you get an incorrect error message:
```
bf luis:convert --in gggg.en-us.lu --out converted.json
Every child entity labelled in an utterance must have its parent labelled in that utterance. Child entity "woofProperty" does not have its parent labelled in utterance "1" for intent "gggg".
```

## To Reproduce
1. Unpack the attached [en-us.zip](https://github.com/microsoft/botframework-cli/files/6031330/en-us.zip)
2. bf luis:convert --in gggg.en-us.lu --out converted.json
3. See error

## Expected behavior
The LU file should be converted. 

## Additional context
The issue is caused by using a hierarchical label before it is defined in a later file.  There was code that was supposed to handle it, but it was incorrect.

[bug]

chrimc62:
Fixes #1128 

The issue was because of file order the merging would drop one of the entities which would then cause validation to fail.  In the code there was a bogus assignment which caused nothing to happen.  The case was triggered by having the definition of a hierarchical property found in a later LU file than where labels were first used.  In the case of form generation if a property definition is after form-global alphabetically then the merge would be incorrect.  In the zip files attached to the bug form-global came before woof-number.

### DIFF
--- a/packages/lu/src/parser/utils/helpers.js
+++ b/packages/lu/src/parser/utils/helpers.js
@@ -354,7 +354,6 @@ const updateModelBasedOnNDepthEntities = function(utterances, entityParentTree)
                 // Is the entity a root entity? 
                 let isRootEntity = parentsForEntity.find(t => t[0] == "$root$")
                 if (isRootEntity === undefined) {
-                    debugger
                     const errorMsg = `Every child entity labelled in an utterance must have its parent labelled in that utterance. Child entity "${entityInUtterance.entity}" does not have its parent labelled in utterance "${utterance.text}" for intent "${utterance.intent}".`;
                     throw (new exception(retCode.errorCode.INVALID_INPUT, errorMsg));
                 }
--- a/packages/lu/src/parser/utils/helpers.js
+++ b/packages/lu/src/parser/utils/helpers.js
@@ -354,6 +354,7 @@ const updateModelBasedOnNDepthEntities = function(utterances, entityParentTree)
                 // Is the entity a root entity? 
                 let isRootEntity = parentsForEntity.find(t => t[0] == "$root$")
                 if (isRootEntity === undefined) {
+                    debugger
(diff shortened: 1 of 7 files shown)


munozemilio (changes requested):
Can we add a test to cover this scenario please? We may refactor the internal parser and having a test would be great.

boydc2014:
@feich-ms can you help double check this one?

packages/lu/src/parser/utils/helpers.js
```diff
@@ -354,6 +354,7 @@ const updateModelBasedOnNDepthEntities = function(utterances, entityParentTree)
                 // Is the entity a root entity? 
                 let isRootEntity = parentsForEntity.find(t => t[0] == "$root$")
                 if (isRootEntity === undefined) {
+                    debugger
```
feich-ms, line 357: >debugger [](start = 20, length = 8)

Is this for check in?

chrimc62: Good catch. I removed that.

feich-ms:
@chrimc62 @munozemilio I further fixed some logic bug when merging intents and entities. Before the fix, intent or entities merging among different lu files will not merge features correctly(Duplicated intents will occur due to this bug, see the removed dups in collated-luis.json). After the fix, intents or entities merging will merge features correctly and no dup intents or entities definitions occur in merged luis json file. I also added/adjusted the test cases to test this fix.

chrimc62:
Looks good to me--thanks!

feich-ms:
@munozemilio could you please help to review?

munozemilio (approved):
I talked to Fei over teams to ask for the modifications on the test

Issue #1082, opened 2020-12-22
LU conversion uses a wrong max number of features


moabba:
## Versions
bf-lu 4.11.1

## Describe the bug
The bf-lu package validates that a model can't have more than 10 features assigned to it. The correct behavior is that a model can have 10 phrase lists features and 10 models as features, totaling 20 features.

## To Reproduce
- Create an application with 10 global features.
- Create an ML entity.
- Add the ML entity as a feature to the "None" intent.
- Export the application as Lu.
- The export will fail with the error "[ERROR] intent None has 11 descriptors (feature). At most 10 is allowed.".

## Expected behavior
A model can have 10 phrase lists features and 10 models as features, totaling 20 features.

[bug]

PR #1131, opened 2021-02-25
Fix max features numbers validation issue

feich-ms:
Fix https://github.com/microsoft/botframework-cli/issues/1082. The bf-lu package validates that a model can't have more than 10 features assigned to it. The correct behavior in luis portal is that a model can have 10 phrase lists features and 10 models as features, totaling 20 features. This fix is to align with the correct behavior in luis portal.


feich-ms:
@munozemilio could you take a look at this PR. This is to fix the maximum features validation issue reported by luis team. More details can be found here https://github.com/microsoft/botframework-cli/issues/1082

boydc2014 (approved):

munozemilio (approved):

Issue #1138, opened 2021-03-02
[bf-lu] Conversion from Lu to JSON fails when a required features has a space in its name


moabba:
## Versions
What CLI version are you using.
bf-lu 4.11.1

What Nodejs version are you using
v10.18.1

What command-line interpreters are you using
Powershell

What OS are you using
Windows

## Describe the bug
When using a required feature in an entity that has spaces in its name, the conversion from Lu to JSON fails.

## To Reproduce
Convert a JSON application to Lu with the following entities
```JSON
"entities": [
  {
    "name": "another entity",
    "features": [
      {
        "modelName": "some entity",
        "isRequired": true
      }
    ]
  },
  {
    "name": "some entity"
  }
]
```
It gets converted to
```
> # Entity definitions

@ some entity "another entity"

@ ml "some entity"
```
When it's converted back to JSON it fails with the error
```
[ERROR] line 17:14 - line 17:30: syntax error: no viable alternative at input ' "another entity"'
[ERROR] line 17:14 - line 17:30: syntax error: invalid input '"another entity"' detected. Expecting one of this - end of file, Whitespace, new line, undefined, model description, QnA definition, Intent definition, undefined, undefined, undefined, undefined, undefined
```

When the model that has the feature is a child entity, e.g.
```JSON
"entities": [
  {
    "name": "some entity"
  },
  {
    "name": "another entity",
    "children": [
      {
        "name": "child entity",
        "features": [
          {
            "modelName": "some entity",
            "isRequired": true
          }
        ]
      }
    ]
  }
]
```
It gets converted to
```
> # Entity definitions

@ ml "some entity"

@ ml "another entity"
    - @ some entity "child entity"
```
But then fails when converted back to JSON with the error
```
[ERROR] line 20:0 - line 20:1: Invalid child entity definition found for "- @ some entity "child entity"". Child definitions must start with '- @' and only include a type, name and optionally one or more usesFeature(s) definition.
```
## Expected behavior
The conversion from Lu to JSON would succeed when the referenced model has a space in its name.

[bug]

PR #1139, opened 2021-03-04
Fix lu converter to handle required feature at top level and whitespaces at child entity name correctly

### LINKED ISSUE #1137: [bf-lu] Non-child entities required features are converted to roles when converting lu to JSON
## Versions
bf-lu 4.11.1

v10.18.1

Powershell

What OS are you using
Windows

## Describe the bug
When a required feature is used in a LUIS Lu file on a non-child entity, it gets translated as a role instead of a required feature.

## To Reproduce
Convert a JSON app to Lu that has the following entities
```json
"entities": [
  {
    "name": "some entity"
  },
  {
    "name": "another entity",
    "features": [
      {
        "modelName": "some entity",
        "isRequired": true
      }
    ]
  }
]
```
It gets converted to this, which makes sense.
```
> # Entity definitions

@ ml "some_entity"

@ some_entity "another entity"
```
But when it's converted back to JSON, it becomes
```json
"entities": [
  {
    "name": "some_entity",
    "roles": [
      "another entity"
    ]
  }
]
```
## Expected behavior
Converting to Lu then back to JSON should return the original JSON file.

[bug]

feich-ms:
Fix #1137 and #1138. The main fixes are listed as below:

- Fix the top level **required("isRequired": true)** feature not converted correctly issue
Before the fix, for the luis json below
```
"entities": [
  {
    "name": "another_entity",
    "features": [
      {
        "modelName": "some_entity",
        "isRequired": true
      }
    ]
  },
  {
    "name": "some_entity"
  }
]
```

It will be converted to below lu format, but the first line below will be recognized as `@ <entity_name> <role>` when converting back to json, so this will cause inconsistency between lu and json converting.
```
@ some_entity another_entity
@ ml some_entity
```

After the fix, the above json will be converted to below. Here `*` appended to feature name is used to identify if it's required or not. When converting this lu format to json, it will use `*` to decide filling the isRequired with true.
```
@ ml another_entity useFeature some_entity*
@ ml some_entity
```

- Fix the whitespace issue in child entity name. This is easy to fix, just apply the `writeEntityName` function that was handling entity name whitespace to ensure quotation marks are added correctly.

New test cases are added to verify the fixes. All existing tests remain unchanged.

### DIFF
--- a/packages/lu/src/parser/lufile/visitor.js
+++ b/packages/lu/src/parser/lufile/visitor.js
@@ -46,12 +46,6 @@ class Visitor {
     static recurselyResolveTokenizedUtterance(tokUtt, entities, errorMsgs, srcUtterance) {
         for (const item of tokUtt) {
             if (item === Object(item)) {
-                let entityName = item.entityName.trim()
-                if (entityName && InvalidCharsInIntentOrEntityName.some(x => entityName.includes(x))) {
-                    errorMsgs.push(`Invalid utterance line, entity name ${entityName} cannot contain any of the following characters: [<, >, *, %, &, :, \\, $]`);
-                    continue;
-                }
-
                 if (item.entityValue === undefined) {
                     // we have a pattern.any entity
                     const patternStr = item.role ? `{${item.entityName}:${item.role}}` : `{${item.entityName}}`
--- a/packages/lu/src/parser/lufile/entitySection.js
+++ b/packages/lu/src/parser/lufile/entitySection.js
@@ -2,7 +2,6 @@ const EntitySectionContext = require('./generated/LUFileParser').LUFileParser.En
 const DiagnosticSeverity = require('./diagnostic').DiagnosticSeverity;
 const BuildDiagnostic = require('./diagnostic').BuildDiagnostic;
 const LUSectionTypes = require('./../utils/enums/lusectiontypes');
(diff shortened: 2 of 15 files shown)


feich-ms:
@munozemilio could you please help to review and verify? Thanks.

feich-ms:
@munozemilio I added validations for entity, intent and phraselist names that could be used as features. They cannot contain special char * in their name definitions(align with luis portal). With this validations, we can make sure feature names with * at the tail can be used to represent the required feature. Some entity name validations are moved to parseFileContents file from lu parser to decouple the luis specific limitations per your suggestions. Pls help to take a look. Thanks.

packages/lu/src/parser/lufile/visitor.js
```diff
@@ -46,12 +46,6 @@ class Visitor {
     static recurselyResolveTokenizedUtterance(tokUtt, entities, errorMsgs, srcUtterance) {
         for (const item of tokUtt) {
             if (item === Object(item)) {
-                let entityName = item.entityName.trim()
```
munozemilio, line 49: This may reduce flexibility for other NLU providers if they want to consume LU format in case they allow those characters in their entity names. On the other hand LU format is already very LUIS oriented.

munozemilio (approved):

PR #1145, opened 2021-03-11
bf luis:build support to republish luis app if users specify a new publishing mode

### LINKED ISSUE #1143: bf luis:build cannot republish luis app to direct version mode if there is no lu content changed
## Versions
4.12.0

## Describe the bug
bf luis:build cannot republish to direct version mode if there is no lu content changed.

## To Reproduce
1. Create a lu file
2. Publish it with production slot mode
3. Republish it with direct version mode
4. You will find the the application in luis portal is still in production slot mode rather than direct version mode

## Expected behavior
The application in luis portal should be direct version mode

## Additional context

[bug]

feich-ms:
Fix #1143. The main fix is to support republishing luis app if users specify a new publishing mode, e.g. users have first published a luis app with directionVersionPublish mode, now they can also publish the same lu content to other modes like Production slot or Staging slot. Before the fix, the republishing logic is completely controlled by content comparisons between current lu content and existing luis portal lu content which means users have no ways to republish to other modes if their lu contents have no changes.

### DIFF
--- a/packages/lu/src/parser/lubuild/builder.ts
+++ b/packages/lu/src/parser/lubuild/builder.ts
@@ -236,11 +236,14 @@ export class Builder {
               const dialogFile = path.join(path.dirname(content.path), `${content.name}.dialog`)
               let recognizer = new Recognizer(content.path, content.name, dialogFile, schema)
 
+              // found existing app endpoints
+              let appEndpoints
               // find if there is a matched name with current app under current authoring key
               if (!recognizer.getAppId()) {
                 for (let app of apps) {
                   if (app.name.toLowerCase() === currentApp.name.toLowerCase()) {
                     recognizer.setAppId(app.id)
+                    appEndpoints = app.endpoints
                     break
                   }
                 }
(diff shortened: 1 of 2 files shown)


boydc2014 (approved):

feich-ms:
@munozemilio could you help to approve this. Composer need this fix to unblock their directVersionPublish feature.

packages/lu/src/parser/lubuild/builder.ts
```diff
@@ -504,4 +507,21 @@ export class Builder {
       app.intents = filteredIntents
     }
   }
+
+  forcePublish(endpoints: any, versionId: string, directVersionPublish: boolean, isStaging: boolean) {
+    let forcePublish = false
+    if (endpoints !== undefined) {
+      if (directVersionPublish && !Object.keys(endpoints).find(version => version.includes(versionId))) {
+        forcePublish = true
```
munozemilio, line 515: if forcePublish is always true, maybe this can be reduced to only setting the variable once

munozemilio (approved):
CIL for a possible simplification

packages/lu/src/parser/lubuild/builder.ts
feich-ms: Thanks for the good catch, I just updated the code piece and made only once variable setting.

PR #1158, opened 2021-03-17
fix build for bug 6429 and snapshot incremental update logic

hcyang:


packages/orchestratorlib/src/predict.ts
```diff
@@ -1289,7 +1289,7 @@ export class OrchestratorPredict {
     const example: Example = Example.newIntentExample(
       this.currentUtterance,
       this.currentIntentLabels);
-    const exampleObejct: any = example.toObject();
+    const exampleObejct: any = example.toAlternateObject();
```
tsuwandy, line 1292: exampleObejct-> exampleObject 😊

packages/orchestratorlib/src/predict.ts
```diff
@@ -1309,7 +1309,7 @@ export class OrchestratorPredict {
     const exampleToRemove: Example = Example.newIntentExample(
       this.currentUtterance,
       this.currentIntentLabels);
-    const exampleObejctToRemove: any = exampleToRemove.toObject();
+    const exampleObejctToRemove: any = exampleToRemove.toAlternateObject();
```
tsuwandy, line 1312: exampleObjectToRemove

packages/orchestratorlib/src/orchestratorhelper.ts
```diff
@@ -811,6 +812,60 @@ export class OrchestratorHelper {
     return false;
   }
 
+  // eslint-disable-next-line max-params
+  static getExampleArrayIntentsEntitiesUtterances(
+    exampleArray: any,
```
tsuwandy, line 817: Should this be 
exampleArray: any[]

tsuwandy (approved):

packages/orchestratorlib/src/predict.ts
hcyang: Fixed!

packages/orchestratorlib/src/predict.ts
hcyang: Typo fixed!

packages/orchestratorlib/src/orchestratorhelper.ts
hcyang: It was returned by the getExamples() ABI as any, even though we know it's an example array.

PR #1161, opened 2021-03-17
Add options to support doing luis and qna cross-train seperately

cosmicshuai:
--inner_dialog: cross-train will run inner dialog training
--intra_dialog: cross-train will run intra dialog training

if both of these flags are absent, the  cross-train will run both inner and intra training


packages/lu/src/parser/cross-train/cross-train.js
```diff
@@ -17,7 +17,7 @@ module.exports = {
    * @param {boolean} verbose verbose to indicate whether log warnings and errors or not when parsing cross-train files.
    * @returns {luResult: any, qnaResult: any} trainedResult of luResult and qnaResult or undefined if no results.
    */
-  train: async function (input, intentName, config, verbose) {
+  train: async function (input, intentName, config, verbose, omitLuis, omitQna) {
```
boydc2014, line 20: It might be good to
rename omitLuis to inner-dialog
rename omitQna to intra-dialog 

and include inner-dialog, intra-dialog into one object called options maybe

cosmicshuai: changed the naming and merged into one object during function call

packages/lu/src/parser/cross-train/cross-train.js
```diff
@@ -17,7 +17,7 @@ module.exports = {
    * @param {boolean} verbose verbose to indicate whether log warnings and errors or not when parsing cross-train files.
    * @returns {luResult: any, qnaResult: any} trainedResult of luResult and qnaResult or undefined if no results.
    */
-  train: async function (input, intentName, config, verbose) {
+  train: async function (input, intentName, config, verbose, enrichDialogOpt) {
```
feich-ms, line 20: add descriptions at function header above

packages/lu/src/parser/cross-train/crossTrainer.js
```diff
@@ -27,7 +27,7 @@ module.exports = {
    * @param {any[]} luContents the lu content array whose element includes path and content
    * @param {any[]} qnaContents the qna content array whose element includes path and content
    * @param {any} configObject cross train config json object
-   * @param {any} options some optional parameters including configId, intentName, verbose, importResolver
+   * @param {any} options some optional parameters including configId, intentName, verbose, importResolver, luOnly
```
feich-ms, line 30: remove this

packages/lu/src/parser/cross-train/crossTrainer.js
```diff
@@ -38,18 +38,19 @@ module.exports = {
       const crossTrainConfig = fileHelper.getConfigObject(
         configObject,
         options.intentName || '_Interruption',
-        options.verbose || true)
+        options.verbose || true,
+        options.enrichDialogOpt)
```
feich-ms, line 42: add a default value for enrichDialogOpt, otherwise it will be breaking changes

packages/lu/test/parser/cross-train/crossTrainer.test.js
```diff
@@ -1036,4 +1036,45 @@ describe('luis:cross training tests among lu and qna contents', () => {
...
+
+    const trainedResult = await crossTrainer.crossTrain(luContentArray, qnaContentArray, crossTrainConfig, options)
+    const luResult = trainedResult.luResult
+    const qnaResult = trainedResult.qnaResult
+
+    let luisContent = luResult.get('main').Content;
+    assert.isTrue(luisContent.includes('# DeferToRecognizer_') === false)
+
+    let qnaContent = qnaResult.get('main').Content;
+    assert.isTrue(qnaContent.includes('# DeferToRecognizer_') === false)
```
feich-ms, line 1078: Add assert for inner dialog cross-train, e.g. assert _Interruption intent

feich-ms: Add test for {inner: false, intra: true}

packages/luis/test/commands/luis/crossTrain.test.ts
```diff
@@ -114,4 +119,19 @@ describe('luis:cross-train tests for lu and qna contents', () => {
...
     })
+
+    test
+    .stdout()
+    .command(['luis:cross-train',
+      '--in', `${path.join(__dirname, './../../fixtures/testcases/interruption5')}`,
+      '--intentName', '_Interruption',
+      '--config', `${path.join(__dirname, './../../fixtures/testcases/interruption5/mapping_rules.json')}`,
+      '--out', './interruptionGen',
+      '--inner_dialog'])
```
feich-ms, line 130: Add test for --intra_dialog

packages/luis/src/commands/luis/cross-train.ts
```diff
@@ -20,13 +20,14 @@ export default class LuisCrossTrain extends Command {
     config: flags.string({description: 'Path to config file of mapping rules'}),
     intentName: flags.string({description: 'Interruption intent name', default: '_Interruption'}),
     force: flags.boolean({char: 'f', description: 'If --out flag is provided with the path to an existing file, overwrites that file', default: false}),
-    log: flags.boolean({description: 'Writes out log messages to console', default: false})
+    log: flags.boolean({description: 'Writes out log messages to console', default: false}),
+    inner_dialog: flags.boolean({description: 'Only do inner dialog cross train', default: false}),
```
feich-ms, line 24: maybe rename to innerDialog and intraDialog?

feich-ms: Please also add these newly added parameter descriptions to README files in luis folder and cli folder

packages/lu/src/parser/cross-train/cross-train.js
boydc2014: trainOpts/trainerOptions

cosmicshuai: renamed to trainingOpt

packages/lu/src/parser/cross-train/crossTrainer.js
cosmicshuai: removed

packages/lu/src/parser/cross-train/crossTrainer.js
cosmicshuai: added

packages/lu/test/parser/cross-train/crossTrainer.test.js
cosmicshuai: added assert and test for  {inner: false, intra: true} case

packages/luis/src/commands/luis/cross-train.ts
cosmicshuai: added

feich-ms:
@munozemilio could you please help to approve this PR since it will be used by composer

packages/luis/README.md
```diff
@@ -425,6 +425,10 @@ OPTIONS
   --intentName=intentName  [default: _Interruption] Interruption intent name
 
   --log                    Writes out log messages to console
+
+  --innerDialog            Only perform the inner dialog cross train
```
feich-ms, line 429: pls also add this to the README in cli package

cosmicshuai: added

packages/luis/test/commands/luis/crossTrain.test.ts
cosmicshuai: added

boydc2014:
cc @scheyal this is the cli part that allows us to do finer-grain control of cross-train. Once this is taken, we can leverage this in Composer side.

feich-ms (approved):

packages/luis/src/commands/luis/cross-train.ts
```diff
@@ -20,13 +20,15 @@ export default class LuisCrossTrain extends Command {
     config: flags.string({description: 'Path to config file of mapping rules'}),
     intentName: flags.string({description: 'Interruption intent name', default: '_Interruption'}),
     force: flags.boolean({char: 'f', description: 'If --out flag is provided with the path to an existing file, overwrites that file', default: false}),
-    log: flags.boolean({description: 'Writes out log messages to console', default: false})
+    log: flags.boolean({description: 'Writes out log messages to console', default: false}),
+    innerDialog: flags.boolean({description: 'Only do inner dialog cross train', default: false}),
```
munozemilio, line 24: From customer usability perspective I have learned that using dashes rather than camel case its more readable. I understand we have a bunch of camelcase flags but  it may be a good idea to switch to inner-dialog, intra-dialog and luis-recognizer-id

packages/luis/src/commands/luis/cross-train.ts
```diff
@@ -39,7 +41,36 @@ export default class LuisCrossTrain extends Command {
         throw new CLIError('Missing cross train config. Please provide config file path by --config.')
       }
 
-      const trainedResult = await crossTrain.train(flags.in, flags.intentName, flags.config, flags.log)
+      if (flags.luRecognizerID !== 'Luis' && flags.luRecognizerID !== 'Orchestrator') {
```
munozemilio, line 44: you can add options to the flag definition.

```
static flags = {
  name: flags.string({
    char: 'n',                    // shorter flag version
    description: 'name to print', // help description for flag
    hidden: false,                // hide from help
    multiple: false,              // allow setting this flag multiple times
    env: 'MY_NAME',               // default to value of environment variable
    options: ['a', 'b'],          // only allow the value to be from a discrete set
    parse: input => 'output',     // instead of the user input, return a different value
    default: 'world',             // default value if flag not passed (can be a function that returns a string or undefined)
    required: false,              // make flag required (this is not common and you should probably use an argument instead)
    dependsOn: ['extra-flag'],    // this flag requires another flag
    exclusive: ['extra-flag'],    // this flag cannot be specified alongside this other flag
  }),
```

packages/luis/src/commands/luis/cross-train.ts
```diff
@@ -39,7 +41,36 @@ export default class LuisCrossTrain extends Command {
...
+        throw new CLIError('luRecognizerID should be either "Luis" and "Orchestrator".')
+      }
+
+      let trainingOpt = {}
+      if (!flags.innerDialog && !flags.intraDialog) {
+        trainingOpt = {
```
munozemilio, line 50: Since innerDialog and intraDialog have default to false you can set the object only once and void code duplication.

munozemilio (approved):
I left some comments.

munozemilio (changes requested):
CIL

packages/luis/src/commands/luis/cross-train.ts
cosmicshuai: naming changed

packages/luis/src/commands/luis/cross-train.ts
cosmicshuai: added options

packages/luis/src/commands/luis/cross-train.ts
cosmicshuai: remove dup code

boydc2014:
1. I think those two flags should be set to true by default, it should not be a requirement to learn and set those flags in normal case
2. the lu-recognizer-id and qna-recognizer-id, should not apply to the whole cross-train process, it should be applied per dialog, thus, probably it should in the config.json instead of in the options.

packages/cli/README.md
```diff
@@ -839,6 +839,12 @@ OPTIONS
   --intentName=intentName  [default: _Interruption] Interruption intent name
 
   --log                    Writes out log messages to console
+
+  --innerDialog            Only performs the inner dialog cross train
+
+  --intraDialog            Only performs the intra dialog cross train
+
+  --luRecognizerID         Recognizer ID of language understanding, shoule be one of Luis or Orchestrator, default is Luis
```
boydc2014, line 847: cli should not expose luRecognizerID

cosmicshuai: remove lu-recognizer-id

munozemilio (approved):

PR #1162, opened 2021-03-17
Add tests for http errors when getting metaschema

### LINKED ISSUE #1149: SchemaMerger.merge never fails if schema is unreachable
## Versions
`4.11.0-dev.20201025.69cf2b9`
`14.15.4`

## Describe the bug
If the schema that is used is unreachable (i.e. returns a 500), merge never fails. 

## Expected behavior
I expect SchemaMerger to retry to fetch the schema in certain cases and eventually fail.

[bug]

chrimc62:
Fixes #1149
Adds tests to make sure that merge works properly when getting failures.


packages/dialog/src/library/schemaMerger.ts
```diff
@@ -576,14 +577,17 @@ export class SchemaMerger {
...
                     } else if (!this.metaSchema) {
-                        // Pick up meta-schema from first .dialog file
-                        this.metaSchemaId = component.$schema
-                        this.currentFile = this.metaSchemaId
-                        this.metaSchema = await getJSON(component.$schema)
-                        this.validator.addSchema(this.metaSchema, 'componentSchema')
-                        this.vlog(`  Using ${this.metaSchemaId} to define components`)
-                        this.currentFile = path
-                        this.validateSchema(component)
+                        if (!triedMetaSchema) {
```
munozemilio, line 580: Unless there is something that needs to be done if the first statement is true, regardless of the second, I'd go with both in the same line as it is clearer that both statements need to be true for something to happen. Also to avoid unnecessarily indent
```
(!this.metaSchema &&
!triedMetaSchema)
```

munozemilio (approved):
Small comment

PR #1170, opened 2021-03-31
Fix bug with npm scoped packages

chrimc62:
Scoped packages like @microsoft/foo would have all of their imports deleted because the scope was considered a package which does not exist.  
Also cleaned-up negative patterns by resolving and by ensuring that the output .schema and .uischema were deleted.


joshgummersall (approved):
Couple quick comments, otherwise LGTM.

munozemilio (approved):

PR #1185, opened 2021-04-07
Ignore multiple copies of the same file

### LINKED ISSUE #1184: Dialog Merge should ignore duplicate files that are exactly the same
If you fail to include the below highlighted line in your packages nuspec file you will get duplicate .schema & .uischema files written to disk. This causes Dialog Merge to fail because of the detected duplicates. Dialog Merge should just ignore duplicate files. 

image

chrimc62:
Fixes #1184 

Treats files with identical content as not being duplicated.  This is because
nuget likes to copy the same files in multiple places.


munozemilio (changes requested):
Can we add a test to cover this case please?

chrimc62:
There already is--that is why the duplicated files were added.  They don't trigger errors which is the test.

---
In reply to: [630774215](https://github.com/microsoft/botframework-cli/pull/1185#issuecomment-630774215) [](ancestors = 630774215)

munozemilio (approved):

PR #1190, opened 2021-04-13
Switch to axios to support proxy server when merging.

chrimc62:
Normalize sorting to reduce variation.


boydc2014 (approved):

packages/dialog/src/library/getJSON.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import * as fs from 'fs-extra'
+const axios = require('axios')
+const axiosHttpsProxy = require('axios-https-proxy')
+axios.interceptors.request.use(axiosHttpsProxy)
+
+const filePrefix = 'file:///'
+
+// Get JSON from a URL.
+export default async function getJSON(url: string): Promise<any> {
```
joshgummersall, line 14: `async function getJSON<T = any>(url: string): Promise<T> { ... }` allows some flexibility without forcing everyone to accept `any`. You may have to add `return resp.data as T` as well.

packages/dialog/src/library/getJSON.ts
```diff
@@ -0,0 +1,21 @@
+/*!
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import * as fs from 'fs-extra'
+const axios = require('axios')
```
joshgummersall, line 7: It seems odd to mix and match import/require. Is there a convention here @munozemilio?

packages/dialog/package.json
```diff
@@ -45,10 +45,11 @@
     "@types/lru-cache": "^5.1.0",
     "@types/xml2js": "^0.4.4",
     "ajv": "^6.12.2",
+    "axios": "~0.21.1",
+    "axios-https-proxy": "^0.1.1",
```
munozemilio, line 49: There is a warning thrown for this: 
seems there is a warning coming from the latest changes in LU: npm WARN axios-https-proxy@0.1.1 requires a peer of axios@^0.18.0 but none is installed. You must install peer dependencies yourself.

boydc2014: Synced will @cosmicshuai, we had a cleaner way to remove this dependency and the warning. @cosmicshuai how about you commit to this PR for this?

munozemilio (changes requested):
CIL

cosmicshuai:
@munozemilio I modified the code, and removed the dependency of `axios-https-proxy`. It will no longer throwing this warning.
I tested the luis publish command via a https proxy.
image

packages/dialog/package.json
cosmicshuai: removed `axios-https-proxy` and commit a change to handle the https proxy setting locally.

packages/dialog/src/library/getJSON.ts
chrimc62: The purpose of this function is to return JSON.  Why do we need the flexibility?  

---
In reply to: [612850256](https://github.com/microsoft/botframework-cli/pull/1190#discussion_r612850256) [](ancestors = 612850256)

packages/dialog/src/library/getJSON.ts
chrimc62: Generally speaking I do import when TS and otherwise do require.  Happy if there is a better pattern.

---
In reply to: [612850391](https://github.com/microsoft/botframework-cli/pull/1190#discussion_r612850391) [](ancestors = 612850391)

munozemilio:
> @munozemilio I modified the code, and removed the dependency of `axios-https-proxy`. It will no longer throwing this warning.
> I tested the luis publish command via a https proxy.
> image

Where is the "Accessing non-existing property" warning coming from?

boydc2014:
> > @munozemilio I modified the code, and removed the dependency of `axios-https-proxy`. It will no longer throwing this warning.
> > I tested the luis publish command via a https proxy.
> > image
> 
> @cosmicshuai Where is the "Accessing non-existing property" warning coming from?

That's comming from anltr on node14, which Shuai is on.

munozemilio (approved):

Issue #1205, opened 2021-04-21
bf luis:cross-train imports don't work correctly


gabog:
## Versions
@microsoft/botframework-cli/4.14.0-dev.20210416.ab0fc09 win32-x64 node-v14.15.4

## Describe the bug
Running bf luis:crosstrain on a composer project that has qna dialogs with imports does not generate the proper dialogs

## To Reproduce
1. Open composer and create a basic assistant bot (don't run or configure the bot)
![image](https://user-images.githubusercontent.com/12264946/115557227-83c1e200-a27f-11eb-80cd-b8f6806f7f80.png)
2. Close composer, open a command prompt and navigate to source folder
3. Run the following commands (ensure the bin and obj folders are deleted)
```
rmdir bin
rmdir obj
mkdir generated/interruption
bf luis:cross-train --in . --out generated/interruption --config settings/cross-train.config.json --force
```
4. Go to the generated/interruption folder and note the following:
![image](https://user-images.githubusercontent.com/12264946/115558468-c637ee80-a280-11eb-9062-4a154b348d1b.png)
  * There is a chitchat_professional.source.en-us.qna file (that shouldn't be there)
  * ChitchatDialog.en-us.qna is empty (it should contain the content from chitchat_professional.source.en-us.qna
  
![image](https://user-images.githubusercontent.com/12264946/115558601-e7004400-a280-11eb-8744-2b6e5bca066e.png)

## Expected behavior
luis:cross-train should have resolved the import in dialogs\ChitchatDialog\knowledge-base\en-us\ChitchatDialog.en-us.qna and included that content into the generated folder file and it didn't.
![image](https://user-images.githubusercontent.com/12264946/115559441-b8369d80-a281-11eb-9acb-cab73389b2fb.png)

Note also that the bin folder also contains copies of the generated folder and that also cause problems in the cross-train operation (we should probably exclude the bin and obj folders from cross-train)

PR #1209, opened 2021-04-22
fix luis-cross train does not find imported files with locale

cosmicshuai:
closes: #1205 

In this PR, the import resolver for luis and qna files will find the correct import files based on the locale info.
Suppose we have a.en-us.lu with content
```
[import](b.lu)
```
The import resolver will find the b.en-us.lu


packages/lu/src/parser/cross-train/cross-train.js
```diff
@@ -19,14 +19,23 @@ module.exports = {
...
-    const qnaContents = await filehelper.getFilesContent(input, fileExtEnum.QnAFile)
+    const luContents = await filehelper.getFilesContent(input, fileExtEnum.LUFile, ignoredFolders)
+    const qnaContents = await filehelper.getFilesContent(input, fileExtEnum.QnAFile, ignoredFolders)
     const configContent = await filehelper.getConfigContent(config)
+    const defaultLocale = 'en-us'
 
     let importResolver = async function (id, idsToFind) {
       let importedContents = []
+      const idWithoutExt = path.basename(id, path.extname(id))
+      const locale = /\w\.\w/.test(idWithoutExt) ? idWithoutExt.split('.').pop() : defaultLocale;
```
feich-ms, line 38: >const locale = /\w\.\w/.test(idWithoutExt) ? idWithoutExt.split('.').pop() : defaultLocale; [](http://example.com/codeflow?start=6&length=91)

This is not always true, for instance, a.xxx.lu, your logic here will make locale be xxx.

packages/luis/test/fixtures/testcases/testImportWithLocale/bin/assistant_core.en-us.lu
```diff
@@ -0,0 +1,43 @@
+# BotTour
+- bot tour
+- onboarding
+- get started
+- start the tour
+- what can you do
+- what can i ask you
```
feich-ms, line 7: I think just one file is enough

feich-ms (approved):
:shipit:

packages/luis/test/fixtures/testcases/testImportWithLocale/bin/assistant_core.en-us.lu
cosmicshuai: modified

boydc2014 (approved):

packages/cli/README.md
```diff
@@ -891,6 +891,8 @@ OPTIONS
   --[no-]intra-dialog      Only do intra dialog cross train
 
   --log                    Writes out log messages to console
+
+  --ignore                ignore given folders under the input directory, for example, --ignore bin,obj,lib, this will ignore the /bin, /obj, /lib folders under the input path
```
gabog, line 895: can we name the paramater exclude instead of ignore

boydc2014:
@munozemilio this is a fix for adaptive runtime CI/CD, it's been verified, can you help get it in

packages/cli/README.md
cosmicshuai: right, that's more accurate. I changed it to `exclude`

gabog, line 895: Hi, I think you missed the readmes (I see ignore is still being used here and in other places)

gabog (dismissed):
I think you missed changing the ignore parameter in the readme files (I still see --ignore there and should be --exclude, right?)

packages/cli/README.md
cosmicshuai: right. I've changed the parameter name in two readme files

packages/luis/src/commands/luis/cross-train.ts
```diff
@@ -22,7 +22,8 @@ export default class LuisCrossTrain extends Command {
     force: flags.boolean({char: 'f', description: 'If --out flag is provided with the path to an existing file, overwrites that file', default: false}),
     log: flags.boolean({description: 'Writes out log messages to console', default: false}),
     'inner-dialog': flags.boolean({description: 'Only do inner dialog cross train', default: true, allowNo: true}),
-    'intra-dialog': flags.boolean({description: 'Only do intra dialog cross train', default: true, allowNo: true})
+    'intra-dialog': flags.boolean({description: 'Only do intra dialog cross train', default: true, allowNo: true}),
+    exclude: flags.string({description: 'Excludes folders under the input directory, seperated by ",". If not specified, all luis and qna files will be included in the cross-train'})
```
munozemilio, line 26: typo: 'separated'

packages/luis/README.md
```diff
@@ -431,6 +431,8 @@ OPTIONS
   --[no-]intra-dialog      Only do intra dialog cross train
 
   --log                    Writes out log messages to console
+
+  --excluder                Excludes given folders under the input directory, for example, --exclude bin,obj,lib, this will ignore the /bin, /obj, /lib folders under the input path
```
munozemilio, line 435: exclude ?

munozemilio (changes requested):
Should content be added to the test empty files? Some typos also need to be addressed.

packages/luis/src/commands/luis/cross-train.ts
cosmicshuai: fixed

packages/luis/README.md
cosmicshuai: fixed

cosmicshuai:
@munozemilio The empty content test files are from a real Composer bot, I think it is OK to keep them empty. There are import files with locale. So the test covers the code added in this PR.

packages/luis/README.md
```diff
@@ -431,6 +431,8 @@ OPTIONS
   --[no-]intra-dialog      Only do intra dialog cross train
 
   --log                    Writes out log messages to console
+
+  --exclude                Excludes given folders under the input directory, for example, --exclude bin,obj,lib, this will ignore the /bin, /obj, /lib folders under the input path
```
munozemilio, line 435: FYI This will be overriden by the doc script.

munozemilio (approved):

PR #1219, opened 2021-04-26
Replace final parser.dereference to speed up merges (#1203)

munozemilio:
* Start exploring schema verification.

* Verify $ref as part of verifySchema
This makes merging much faster.

Co-authored-by: Chris McConnell <chrimc>

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -58,7 +58,7 @@
         "999999",
         "--colors",
         "-g",
-        "dialog:merge.*"
+        "dialog:merge.*missing schema.*"
       ],
       "cwd": "${workspaceFolder}/packages/dialog",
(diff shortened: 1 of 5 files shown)


.vscode/launch.json
```diff
@@ -58,7 +58,7 @@
         "999999",
         "--colors",
         "-g",
-        "dialog:merge.*"
+        "dialog:merge.*missing schema.*"
```
chrimc62, line 61: >      "dialog:merge.*missing schema.*" [](http://example.com/codeflow?start=2&length=38)

This should not be there, but is OK since this is just for dev testing.

chrimc62 (approved):
:shipit:

joshgummersall (approved):

PR #1229, opened 2021-05-06
Support extracting more modelTypeId in luis:test command

cosmicshuai:
currently, in luis:test, only those entities with modelTypeId is 1 or 4, or those entities have roles will be included. 
https://github.com/microsoft/botframework-cli/blob/968ef754d04a5096e8471ba12fed5db9cc853606/packages/lu/src/parser/test/testhelper.ts#L108
In this PR, all modelTypeId  1 or 4 or 10, or those entities have roles will be included.


munozemilio (approved):

munozemilio:
Should we add a test?

cosmicshuai:
> Should we add a test?

I slightly modified a test to cover the case when the modelTypeId is 10

PR #1246, opened 2021-06-23
use 4..14.0-preview version of orchestrator-core

tsuwandy:
Issue #1247

### DIFF
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -3,6 +3,7 @@ dependencies:
   '@azure/cognitiveservices-luis-authoring': 4.0.0-preview.1
   '@azure/cognitiveservices-luis-runtime': 5.0.0
   '@azure/ms-rest-azure-js': 2.0.1
+  '@microsoft/orchestrator-core': 4.14.0-preview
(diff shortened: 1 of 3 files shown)


common/config/rush/pnpm-lock.yaml
```diff
@@ -318,6 +318,25 @@ packages:
...
+      - x64
+      - ia32
+    dependencies:
+      '@mapbox/node-pre-gyp': 1.0.3
+      bindings: 1.2.1
+      node-addon-api: 3.1.0
+      node-gyp: 8.0.0
+    dev: false
+    engines:
+      node: ^10.0.0 || ^12.0.0 || ^13.0.0 || ^14.0.0 || ^16.0.0
```
daveta, line 332: Wondering if we should advertise nodejs16 at this point since it's not LTS.

daveta (approved):

common/config/rush/pnpm-lock.yaml
daveta: Looks like they will most likely guard at the top..

taicchoumsft (approved):

munozemilio (approved):

PR #1249, opened 2021-06-24
Changes for R14

tsuwandy:
Issue #1252 
- Incremental update for bf orchestrator:create  (if existing blu file exists)
- Added bf orchestrator:remove command
- Added download progress info for Orchestrator base model download onProgress callback


packages/orchestratorlib/src/basemodel.ts
```diff
@@ -100,13 +100,25 @@ export class OrchestratorBaseModel {
...
 
-  public static async downloadModel(modelUrl: string, modelZipPath: string): Promise<(resolve: any) => void> {
+  public static async downloadModel(modelUrl: string, modelZipPath: string, onProgress: any = OrchestratorBaseModel.defaultHandler): Promise<(resolve: any) => void> {
     const response: AxiosResponse<Stream> = await axios({
       method: 'GET',
       url: modelUrl,
       responseType: 'stream',
     });
+    const totalLength: number = response.headers['content-length'];
+    onProgress(`Total to download: ${totalLength} bytes...`);
```
daveta, line 110: Neat.

daveta (approved):

packages/orchestrator/src/commands/orchestrator/add.ts
```diff
@@ -57,33 +57,32 @@ export default class OrchestratorAdd extends Command {
         fullEmbeddings = true;
       }
 
+      const settings: OrchestratorSettings = OrchestratorSettings.getCurrent();
```
joshgummersall, line 60: Is the return type of `getCurrent` already `OrchestratorSettings`? Generally, I prefer to omit type annotations that are automatically inferred by the compiler. You should be able to write:
```typescript
const settings = OrchestratorSettings.getCurrent();
```

joshgummersall: I noticed this in a lot of places so I won't comment on all of them, it's definitely something that could get cleaned up if there's time.

packages/orchestrator/src/commands/orchestrator/create.ts
```diff
@@ -55,45 +56,45 @@ export default class OrchestratorCreate extends Command {
 
       if (DataSourceHelper.isDispatchInput(input)) {
         const dispatchJson: any = JSON.parse(OrchestratorHelper.readFile(input));
-        DataSourceHelper.convertDispatchInputs(dispatchJson, OrchestratorSettings.DataSources);
+        DataSourceHelper.convertDispatchInputs(dispatchJson, settings.DataSources);
         refresh = true;
       }
 
-      const hasDataSources: boolean = OrchestratorSettings.DataSources && OrchestratorSettings.DataSources.inputs.length > 0;
+      const hasDataSources: boolean = settings.DataSources && settings.DataSources.inputs.length > 0;
```
joshgummersall, line 63: You can write this as `const hasDataSources = settings.DataSources?.inputs?.length > 0 ?? false;`

packages/orchestrator/src/commands/orchestrator/create.ts
```diff
@@ -55,45 +56,45 @@ export default class OrchestratorCreate extends Command {
...
+      settings.persist();
     } catch (error) {
       throw (new CLIError(error));
     }
   }
 
-  private async refreshLuisQnAInputs(dataSources: OrchestratorDataSourceSettings): Promise<void> {
-    if (!dataSources) {
+  private async refreshLuisQnAInputs(settings: OrchestratorSettings): Promise<void> {
+    if (!settings || !settings.DataSources) {
```
joshgummersall, line 89: `if (!settings?.Datasources) {}`

packages/orchestrator/src/commands/orchestrator/remove.ts
```diff
@@ -0,0 +1,58 @@
...
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ * Licensed under the MIT License.
+ */
+
+import * as path from 'path';
+import {Command, CLIError, flags} from '@microsoft/bf-cli-command';
+import {Orchestrator, OrchestratorDataSource, OrchestratorSettings} from '@microsoft/bf-orchestrator';
+
+export default class OrchestratorRemove extends Command {
+  static description: string = 'Remove examples from LUIS app(s), QnaMaker kb(s) or .lu/.qna/.json files from Orchestrator snapshot file.';
```
joshgummersall, line 11: `: string` is inferred.

packages/orchestrator/src/commands/orchestrator/remove.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  static examples: Array<string> = [`
```
joshgummersall, line 13: `: Array<string>` is inferred. Is this meant to be a list of separate examples, or a single item list of one multiline example?

packages/orchestrator/src/commands/orchestrator/remove.ts
```diff
@@ -0,0 +1,58 @@
...
+        } else {
+          this.log(`Removed ${dataSource.Type} source with id ${dataSource.Id}`);
+        }
+      } else {
+        this.log('Data source not found');
+      }
+
+      settings.persist();
+    } catch (error) {
+      throw (new CLIError(error));
```
joshgummersall, line 55: Parens around `new CLIError` are unnecessary.

packages/orchestratorlib/src/build.ts
```diff
@@ -279,6 +275,12 @@ export class OrchestratorBuild {
     });
   }
 
+  // Synchronize an active LabelResolver instance with an LU file.
+  static async syncLabelResolver(labelResolver: LabelResolver, luContent: string) {
+    const target: ITextUtteranceLabelMapDataStructure = await OrchestratorBuild.getExamplesLU(luContent);
+    OrchestratorBuild.syncLabelResolverEx(labelResolver, target);
```
joshgummersall, line 281: You should either `await` this function, too, or `return OrchestratorBuild.syncLabelResolverEx(...)` so the promise is propagated.

packages/orchestratorlib/src/create.ts
```diff
@@ -11,6 +11,7 @@ import {ITextUtteranceLabelMapDataStructure} from '@microsoft/bf-dispatcher';
 // import {Label} from '@microsoft/bf-dispatcher';
 
 import {LabelResolver} from './labelresolver';
+import {OrchestratorBuild} from '.';
```
joshgummersall, line 14: Generally, I think it's better to use the file path here rather than implicitly referencing `index.ts`.

packages/orchestratorlib/src/create.ts
```diff
@@ -45,42 +46,16 @@ export class OrchestratorCreate {
...
-     *    } else {
-     *      Utility.debuggingLog('');
-     *    }
-     *    for (const label of labels) {
-     *      Utility.debuggingLog(`label=${label.name}`);
-     *    }
-     *  }
-     */
+
+    OrchestratorBuild.syncLabelResolverEx(labelResolver, processedUtteranceLabelsMap);
```
joshgummersall, line 58: `await`?

packages/orchestratorlib/src/build.ts
```diff
@@ -154,14 +151,13 @@ export class OrchestratorBuild {
     return luIntentsEntitiesUtterances;
   }
 
-  // Synchronize an active LabelResolver instance with with an LU file.
-  static async syncLabelResolver(labelResolver: LabelResolver, luContent: string) {
+  // Synchronize an active LabelResolver instance with ITextUtteranceLabelMapDataStructure.
+  static async syncLabelResolverEx(labelResolver: LabelResolver, target: ITextUtteranceLabelMapDataStructure) {
```
joshgummersall, line 155: Also, add `: Promise<void>` as the return type here.

packages/orchestratorlib/src/build.ts
```diff
@@ -279,6 +275,12 @@ export class OrchestratorBuild {
     });
   }
 
+  // Synchronize an active LabelResolver instance with an LU file.
+  static async syncLabelResolver(labelResolver: LabelResolver, luContent: string) {
```
joshgummersall, line 279: Add `: Promise<void>` as the return type.

packages/orchestratorlib/src/settings.ts
```diff
@@ -124,23 +161,33 @@ export class OrchestratorSettings {
...
   }
 
-  public static addUpdateDataSource(data: OrchestratorDataSource) {
+  public addUpdateDataSource(data: OrchestratorDataSource) {
     this.DataSources.inputs.push(data);
   }
 
-  public static ModelPath: string;
+  public static getCurrent(): OrchestratorSettings {
+    if (!OrchestratorSettings.Current) {
```
joshgummersall, line 173: `OrchestratorSettings.Current ??= new OrchestratorSettings();`

packages/orchestratorlib/src/settings.ts
```diff
@@ -152,59 +199,57 @@ export class OrchestratorSettings {
...
-  public static init(
+  public init(
     settingsDir: string,
     baseModelPath: string,
     entityBaseModelPath: string,
     snapshotPath: string,
     hierarchical: boolean = false,
-    hasDataSources: boolean = false)  {
+    hasDataSources: boolean = false,
+    settingsFileName: string = OrchestratorSettings.OrchestratorSettingsFileName)  {
```
joshgummersall, line 209: Add `: void` return type.

packages/orchestratorlib/tsconfig.json
```diff
@@ -9,6 +9,7 @@
     "module": "commonjs",
     "outDir": "./lib",
     "pretty": true,
+    "strictNullChecks": false,
```
joshgummersall, line 12: `strictNullChecks` is a wonderful feature, why is this necessary?

taicchoumsft (approved):
:shipit:

packages/orchestrator/src/commands/orchestrator/add.ts
tsuwandy: yeah, it's annoying but we use eslint and it complains if I don't have the type in there... :-(

packages/orchestrator/src/commands/orchestrator/remove.ts
tsuwandy: unfortunately, eslint is forcing this. :-(

packages/orchestrator/src/commands/orchestrator/remove.ts
tsuwandy: same deal with the type annotation, thanks to the annoying eslint. :-)  

image

packages/orchestratorlib/src/build.ts
tsuwandy: fixed!

packages/orchestratorlib/src/settings.ts
tsuwandy: ok :-)

packages/orchestratorlib/tsconfig.json
tsuwandy: ah, i wanted https://github.com/microsoft/botframework-cli/blob/8cae60197f53eab4dc93b4f5d63001356c811755/packages/orchestratorlib/src/settings.ts#L113 to return null and the caller to check for null.  Wasn't possible without it.  Thoughts on alternative?

packages/orchestratorlib/src/settings.ts
tsuwandy: uh, unfortunately eslint doesn't like OrchestratorSettings.Current ??= new OrchestratorSettings();

D:\src\botframework-cli\packages\orchestratorlib>npm run posttest

> @microsoft/bf-orchestrator@1.0.0 posttest D:\src\botframework-cli\packages\orchestratorlib
> eslint . --ext .ts --config .eslintrc

TypeError: Cannot read property 'range' of null
Occurred while linting D:\src\botframework-cli\packages\orchestratorlib\src\settings.ts:173
    at SourceCode.getTokenBefore (D:\src\botframework-cli\common\temp\node_modules\.pnpm\registry.npmjs.org\eslint\5.16.0\node_

reverting change.. :-(

packages/orchestratorlib/tsconfig.json
tsuwandy: ok, will use union type instead and remove "strictNullChecks": false,

packages/orchestrator/src/commands/orchestrator/remove.ts
munozemilio: It would be nice to filter expected errors from Orchestrator otherwise telemetry will never catch unexpected issues

packages/orchestratorlib/src/settings.ts
```diff
@@ -152,59 +198,57 @@ export class OrchestratorSettings {
...
+
+      this.SnapshotPath = OrchestratorSettings.ensureSnapshotPath(snapshotPath, settingsDir, settings);
+      this.ModelPath = OrchestratorSettings.ensureBaseModelPath(baseModelPath, settings);
       entityBaseModelPath = OrchestratorSettings.ensureEntityBaseModelPath(entityBaseModelPath, settings);
       if (!Utility.isEmptyString(entityBaseModelPath)) {
-        OrchestratorSettings.EntityModelPath = entityBaseModelPath;
+        this.EntityModelPath = entityBaseModelPath;
       }
-    }  catch {}
+    } catch {}
```
munozemilio, line 232: Whats the purpose of an empty catch?

packages/orchestrator/src/commands/orchestrator/basemodel/get.ts
```diff
@@ -64,10 +66,15 @@ export default class OrchestratorBaseModelGet extends Command {
...
             modelInfo.modelFolder,
             modelInfo.versionId,
-            (message: any) => {
-              if (flags.verbose) {
+            (message: any, pct: number) => {
+              if (message) {
                 this.log(message);
               }
+              if (flags.verbose) {
+                if (pct) {
```
munozemilio, line 74: Nit
```
if (flags.verbose
    && pct) { 
    ....
}
```
This could reduce indentation a bit.

munozemilio (approved):
I left some small comments.

packages/orchestrator/src/commands/orchestrator/basemodel/get.ts
tsuwandy: makes sense! :-)

PR #1266, opened 2021-07-07
syncLabelResolverEx should return after addBatch

### LINKED ISSUE #1262: Don't do incremental update to LabelResolver when there is no existing blu file
Doing incremental update is way slower when LabelResolver.addBatch when there is no existing blu file.  This is a regression from 4.13 version for when there is no existing snapshot.

tsuwandy:
Fix #1262

### DIFF
--- a/packages/orchestratorlib/src/build.ts
+++ b/packages/orchestratorlib/src/build.ts
@@ -161,6 +161,7 @@ export class OrchestratorBuild {
       Utility.debuggingLog('OrchestratorBuild.syncLabelResolverEx(), ready to call LabelResolver.addBatch()');
       LabelResolver.addBatch(target);
       Utility.debuggingLog('OrchestratorBuild.syncLabelResolverEx(), after calling LabelResolver.addBatch()');
+      return;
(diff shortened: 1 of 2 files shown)


packages/orchestratorlib/src/build.ts
```diff
@@ -161,6 +161,7 @@ export class OrchestratorBuild {
       Utility.debuggingLog('OrchestratorBuild.syncLabelResolverEx(), ready to call LabelResolver.addBatch()');
       LabelResolver.addBatch(target);
       Utility.debuggingLog('OrchestratorBuild.syncLabelResolverEx(), after calling LabelResolver.addBatch()');
+      return;
```
munozemilio, line 164: Should we add a test to catch this scenario?

daveta (approved):

hcyang (approved):

munozemilio (changes requested):
If this was not caught by the tests, maybe we should add one

munozemilio (approved):

PR #1278, opened 2021-07-30
add children entities

hcyang:
Process bf-cli parsed JSON and consume children entity labels.

### DIFF
--- a/packages/orchestratorlib/src/orchestratorhelper.ts
+++ b/packages/orchestratorlib/src/orchestratorhelper.ts
@@ -75,10 +75,6 @@ export class OrchestratorHelper {
     fs.unlinkSync(filePath);
   }
 
-  public static label(name: string, offset: number, length: number): {entity: string; startPos: number; endPos: number} {
-    return {entity: name, startPos: offset, endPos: offset + length - 1};
-  }
-
   public static createDteContent(utteranceLabelsMap: Map<string, Set<string>>) {
     const labelUtteranceMap: Map<string, string> = new Map<string, string>();
     // eslint-disable-next-line guard-for-in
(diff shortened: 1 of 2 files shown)


packages/orchestratorlib/test/orchestratorhelper.test.ts
```diff
@@ -352,6 +353,78 @@ describe('Test Suite - orchestratorhelper', () => {
     assert.ok((utteranceEntityLabelDuplicateMap.get(utterance1) as Label[]).length === 3,
       `(utteranceEntityLabelDuplicateMap.get(utterance1) as Label[]).length=${(utteranceEntityLabelDuplicateMap.get(utterance1) as Label[]).length}`);
   });
+  it('Test.0201 OrchestratorHelper.getJsonIntentsEntitiesUtterances() with child entity labels', function () {
+    Utility.resetFlagToPrintDebuggingLogToConsole(UnitTestHelper.getDefaultUnitTestDebuggingLogFlag());
+    this.timeout(UnitTestHelper.getDefaultUnitTestTimeout());
+    const utterance: string = 'traveling from Paris to Berlin';
+    const jsonObjectArray: {
+      'text': string;
+      'intents': string[];
```
daveta, line 362: Curious can intent have start and end pos.

hcyang: Not in the LU/LUIS JSON format.

packages/orchestratorlib/test/orchestratorhelper.test.ts
```diff
@@ -352,6 +353,78 @@ describe('Test Suite - orchestratorhelper', () => {
...
+      'entities': {
+        'entity': string;
+        'startPos': number;
+        'endPos': number;
+        'children': {
```
daveta, line 367: Assuming all children span be contained within parent span? And are there different child types in future?

hcyang: I do hope so. The bf-lu parser produces the JSON file with children entities completely encompassed within a parent entity's span. Cannot predict the future :), Just try to deal with what's defined today.

packages/orchestratorlib/src/orchestratorhelper.ts
```diff
@@ -859,9 +856,10 @@ export class OrchestratorHelper {
                 } else if (labelType === LabelType.Entity) {
                   const labelSpanOffset: number = label.span.offset;
                   const labelSpanLength: number = label.span.length;
-                  OrchestratorHelper.addNewEntityLabelUtterance(
+                  const newEntityLabel: Label = Label.newEntityLabel(labelName, labelSpanOffset, labelSpanLength);
```
chrimc62, line 859: >newEntityLabel [](http://example.com/codeflow?start=54&length=14)

This will not work unless you map somewhere else.  The reason the local label function was introduced is because label.newEntityLabel was not consistent with what the parser emitted.  Try running through creation twice.

chrimc62 (changes requested):
:clock1:

packages/orchestratorlib/src/orchestratorhelper.ts
hcyang: > > newEntityLabel
> 
> This will not work unless you map somewhere else. The reason the local label function was introduced is because label.newEntityLabel was not consistent with what the parser emitted. Try running through creation twice.

@chrimc62, Orchestrator-Core internally stores entity labels in offset/length data structure, instead of LUIS JSON's startPos/endPos. The getExampleArrayIntentsEntitiesUtterances() function was used to consume an "example" array emitted by Orchestrator-Core, so the input it got are in the offset/length format. It was a bug (that you fixed before) of using addNewEntityLabelUtterance(), which was designed for consuming LUIS JSON's startPos/endPos labels.

This time around, addNewEntityLabelObjectUtterance() was used and it does expect a more "strong-typed" Label object, which is offset/length-based.

chrimc62 (approved):
:shipit:

munozemilio (approved):

PR #1288, opened 2021-09-01
Hide qna endpoint keys from qnamaker:build output, unless --logKeys flag is specified

alexstojda:
EndpointKeys should not be shown in the logs by default as anyone with the key can then make requests to the endpoint.


alexstojda:
Project seems to be abandoned, so I'll close this PR...

Issue #1276, opened 2021-07-26
Undefined error when parsing QnA file using LU parser


chrimc62:
## Versions
@microsoft/botframework-cli/4.15.0-dev.20210723.bf1f054 win32-x64 node-v14.16.0

## Describe the bug
This was reported by a customer. If you have a QnA file with a link across multiple lines then you will get an 'undefined' error rather than any useful diagnostics.  

## To Reproduce
1. Create a QnA File foo.qna like: 
~~~
**Prompts:**
- [Files that Cannot Sync
](#1)

<a id = "1"/>

## ? Files that Cannot Sync
```
Files that cannot sync.
```
~~~
2. bf orchestrator:create -i foo.qna -m basemodel
3. You will get an undefined error in `BuildDialognostic` because context is a string and not a structure. 
```
    else if (context !== undefined) {
        const startPosition = new Position(context.start.line, context.start.column);
```

## Expected behavior
An error message indicating the line.

[bug]

jongon:
I have the same problem here. I'm working with tsv files meanwhile

tsuwandy:
@jongon, could you share the problematic .qna file?

tsuwandy:
Looks like the content in the example above is invalid, see this page for valid .qna content.  https://docs.microsoft.com/en-us/azure/bot-service/file-format/bot-builder-qna-file-format?view=azure-bot-service-4.0

Prompts should be part of a question.  The following will work fine:
# ?store hours
```
Most our stores are open M-F 9AM-10PM.
```
**Prompts:**
- [Files that Cannot Sync](#1)

<a id = "1"></a>

# ? Files that Cannot Sync
```
Files that cannot sync.
```

tsuwandy:
Related issue https://github.com/microsoft/botframework-cli/issues/1294

tsuwandy:
Fix should be available with the next bf cli daily deployment.

fedsp:
Hi @tsuwandy, what do you mean by the daily deployment?
I'm facing the exact same issue using the latest release.

tsuwandy:
@fedsp, yes, it should be in the stable version before end of the month.

PR #1291, opened 2021-09-15
Surface lu parser error 

tsuwandy:
Issue #1276 

Surface error from lu parser with problematic line #.

### DIFF
--- a/packages/orchestratorlib/src/orchestratorhelper.ts
+++ b/packages/orchestratorlib/src/orchestratorhelper.ts
@@ -426,12 +426,10 @@ export class OrchestratorHelper {
         content: luContent,
         id: luFile,
       };
-      
       const luisObject: any = await LuisBuilder.fromLUAsync([luObject], OrchestratorHelper.findLuFiles);
       if (Utility.toPrintDetailedDebuggingLogToConsole) {
         UtilityDispatcher.debuggingNamedLog1('OrchestratorHelper.parseLuContent(): calling getIntentsEntitiesUtterances()', luisObject, 'luisObject');
-      }  
-
+      }
       const rvLu: boolean = OrchestratorHelper.getIntentsEntitiesUtterances(
         luisObject,
         hierarchicalLabel,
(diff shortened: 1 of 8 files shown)


packages/orchestratorlib/src/orchestratorhelper.ts
```diff
@@ -440,9 +440,9 @@ export class OrchestratorHelper {
       if (!rvLu) {
         throw new Error('Failed to parse LUIS or JSON file on intent/entity labels');
       }
-    } catch (error) {
-      Utility.debuggingLog(`EXCEPTION calling getIntentsEntitiesUtterances(), error=${error}`);
-      throw error;
+    } catch (error: any) {
+      const errorText: string = 'text' in error ? error.text : '';
```
chrimc62, line 444: >const errorText: string = 'text' in error ? error.text : ''; [](http://example.com/codeflow?start=6&length=60)

You can get away with just putting error.text in your error string here and in other places.  If it is not present it will be null which will end up the empty string.

chrimc62 (approved):
:shipit:

packages/orchestratorlib/src/orchestratorhelper.ts
```diff
@@ -440,9 +440,8 @@ export class OrchestratorHelper {
       if (!rvLu) {
         throw new Error('Failed to parse LUIS or JSON file on intent/entity labels');
       }
-    } catch (error) {
-      Utility.debuggingLog(`EXCEPTION calling getIntentsEntitiesUtterances(), error=${error}`);
-      throw error;
+    } catch (error: any) {
+        throw new Error(`Failed parsing lu file ${luFile} ${error.text}`);
```
munozemilio, line 444: why text and not message? are you expecting a specific type of Error?

tsuwandy: error.message has nothing but error.text has the line number and error info from the parser.

munozemilio (approved):
Feel free to merge it, but keep in mind runtime errors will be logged as undefined unless bf-lu exceptions are filtered

packages/orchestratorlib/src/orchestratorhelper.ts
```diff
@@ -602,20 +601,19 @@ export class OrchestratorHelper {
...
+          newlines.push(line);
+        }
+      });
+      // Utility.debuggingLog('OrchestratorHelper.parseQnaFile() ready to call QnaMakerBuilder.fromContent()');
+      const qnaNormalized: string = Utility.cleanStringOnTabs(newlines.join('\n')); // ---- NOTE ---- QnaMakerBuilder does not like TAB
+      const qnaObject: any = await QnaMakerBuilder.fromContent(qnaNormalized);
       OrchestratorHelper.getQnaQuestionsAsUtterances(qnaObject, hierarchicalLabel, utteranceLabelsMap, utteranceLabelDuplicateMap);
-    } else {
-      throw new Error(`Failed parsing qna file ${qnaFile}`);
+    } catch (error: any) {
```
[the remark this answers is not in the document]
tsuwandy: We have enough logging everywhere within createasync so should be ok.

